PR #1, opened 2016-08-18
Recent version of redigo have a DialURL

cyberdelia:
Recent version of redigo have a ``DialURL``, which allow us to get rid of ``ParseURL``.


freeformz:
Awesome. I completely missed that!

PR #2, opened 2016-11-16
Updates

freeformz:
* Make this explicitly about redigo.
* Add support for changing the redis:// url to rediss://. This will be
usable once https://github.com/garyburd/redigo/pull/208 lands.
* Remove our own url parser as current versions of redigo already supply
one.


amerine:
LGTM

PR #4, opened 2017-04-10
Extracts heroku/log/midgard/scrubber.go into a 'scrub' package

elight:


scrub/header/header_test.go
```diff
@@ -0,0 +1,71 @@
+package header
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
freeformz: -1

scrub/header/header.go
```diff
@@ -0,0 +1,54 @@
+package header
```
freeformz: Put this in `scrub/header.go`, no need to make a `header` sub package. This would import as header and you would call `header.Header`, which makes no sense really where as `scrub.Header` does make sense.

scrub/url/url.go
```diff
@@ -0,0 +1,86 @@
+package url
```
freeformz: Same for this package.

scrub/url/url_test.go
```diff
@@ -0,0 +1,83 @@
+package url
+
+import (
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
freeformz: Same for these tests.

scrub/header/header_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/stretchr/testify/suite"
```
freeformz: Please use normal go test functions.

scrub/header.go
```diff
@@ -0,0 +1,59 @@
...
+			scrubbedValues := []string{}
+			for _, auth := range v {
+				substrs := strings.SplitN(auth, " ", 2)
+				scrubbed := scrubbedValue
+				if len(substrs) > 1 {
+					scrubbed = fmt.Sprintf("%s %s", substrs[0], scrubbedValue)
+				}
+				scrubbedValues = append(scrubbedValues, scrubbed)
+			}
+			scrubbedHeader[k] = scrubbedValues
```
freeformz: Why not `continue` after this instead of forcing the indent below?

scrub/url.go
```diff
@@ -0,0 +1,85 @@
...
+		"token":                              true,
+		"user_session_secret":                true,
+		"www-sso-session":                    true,
+	}
+)
+
+// URL removes a subset of "sensitive" URL parameters as used by parts of Heroku's
+// conventions for API design. The output of this function is safe to be logged
+// except in high-security scenarios.
+func URL(u *url.URL) *url.URL {
```
freeformz: I  wonder if this should take a string instead of a *url.URL. I guess this is most likely to be used with a http.Request, which includes and already parsed *url.URL though, so nevermind.

scrub/url.go
```diff
@@ -0,0 +1,85 @@
...
+
+		// scrub values that are URLs
+		for _, vv := range v {
+			u, err := url.Parse(vv)
+			if err != nil {
+				continue
+			}
+
+			u.User = scrubURLUserInfo(u.User)
+			query.Set(k, u.String())
```
freeformz: Why not call `query.Set(k, URL(u).String())` and get full sanitization?

scrub/header.go
Xe: Fixed in https://github.com/heroku/x/commit/74b775d3f1beb50f1bf44ea7819cb2f996ce06f3

scrub/url.go
Xe: Fixed in https://github.com/heroku/x/commit/74b775d3f1beb50f1bf44ea7819cb2f996ce06f3

Xe: tests added for this in https://github.com/heroku/x/pull/4/commits/84b13143e8adf8e3dc9b86cf874ecae6d7416fe7

PR #7, opened 2017-05-26
Add optional percentile prefix

cyx:


go-kit/metrics/provider/librato/librato_test.go
```diff
@@ -59,7 +59,7 @@ func TestLibratoReport(t *testing.T) {
 		t.Errorf("got %q but didn't expect any errors", err)
 	}
 
-	p := New(u, time.Second, WithSource("test.source"), WithErrorHandler(errHandler))
+	p := New(u, time.Second, WithSource("test.source"), WithErrorHandler(errHandler), WithPercentilePrefix(defaultPercentilePrefix))
```
freeformz: I would drop this because it's not necessary to get the defaultPercentilPrefix.

cyx: :+1:

PR #8, opened 2017-05-31
submit counters as gauges

cyx:
After chatting with librato folks, it appears they recommend using
gauges for everything.

This PR changes the counters to be submitted as gauges.


freeformz:
:+1:

PR #9, opened 2017-05-31
counters: value reset every submission

cyx:
After trying out the current approach, it appears it's more work
since it will require you to derive every single chart metric.

Doing a reset on the counter simplifies a lot of this.

### DIFF
--- a/go-kit/metrics/provider/librato/librato.go
+++ b/go-kit/metrics/provider/librato/librato.go
@@ -63,6 +63,7 @@ func (e Error) Error() string {
 type Provider struct {
 	errorHandler                     func(err error)
 	source, prefix, percentilePrefix string
+	resetCounters                    bool
 
 	once sync.Once
 	done chan struct{}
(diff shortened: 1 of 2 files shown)


freeformz:
I'm 👎 atm.

From slack:

```
Our most common use of counters is to derive(value()) for a period. I can't think of a use @ heroku where we care about the underived number as we're always looking at periods of time. I guess it's easier to determine arbitrary derivates for any given period when using the absolute count, vs. summing the reported, pre-calculated derivates though. With underived gauges for counters we  could use the, in some cases, to signal restarts and/or other, similar behavior.

Seems like reporting the monotonically increasing values offers more flexibility than reporting the derivatives.

But I could be wrong?
```

cyx:
@freeformz I made it into an option instead. I believe this is a fair compromise, yeah? :)

freeformz:
Yeah, LGTM.

PR #10, opened 2017-06-01
ssa enabled: all metrics are submitted with SSA

cyx:
## Reasons for doing this

Certain projects use a single source for many dynos. This posting style requires SSA enabled for all metrics. To avoid having to manually enable SSA for new metrics created, it's easier to make this automatic.

## Description of changes

- add `WithSSA` option
- submits a payload like the following if enabled:

```
curl -u $LIBRATO_USERNAME:$LIBRATO_TOKEN -H "Content-Type: application/json" -d '{
  "source": "xyz.foobarbaz",
  "gauges": [{
    "name": "xyz.foobarbaz.gauge1",
    "period": 60,
    "min": 1,
    "max": 1,
    "sum": 1,
    "sum_squares": 1,
    "count": 1
  }], "attributes": { "aggregate": true }
}' -X POST 'https://metrics-api.librato.com/v1/metrics' -i
```

### DIFF
--- a/go-kit/metrics/provider/librato/librato.go
+++ b/go-kit/metrics/provider/librato/librato.go
@@ -230,9 +230,19 @@ type gauge struct {
 	Min    float64 `json:"min"`
 	Max    float64 `json:"max"`
 	SumSq  float64 `json:"sum_squares"`
+}
 
-	// Attributes are used to enable SSA for gauges.
-	Attributes map[string]interface{} `json:"attributes,omitempty"`
+// attributes are top level things which you can use to affect newly created
+// metrics.
+type attributes struct {
+	Color             string  `json:"color,omitempty"`
+	DisplayMin        float64 `json:"display_min,omitempty"`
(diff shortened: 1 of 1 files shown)


cyx:
- I verified that creating a new metric this works.
- Also verified that if you submit a pre-existing gauge, plus a new one, that the SSA activation works as well.

go-kit/metrics/provider/librato/librato.go
```diff
@@ -223,6 +230,9 @@ type gauge struct {
 	Min    float64 `json:"min"`
 	Max    float64 `json:"max"`
 	SumSq  float64 `json:"sum_squares"`
+
+	// Attributes are used to enable SSA for gauges.
+	Attributes map[string]interface{} `json:"attributes,omitempty"`
```
Xe: Librato's API docs have a [fixed list of attributes](https://www.librato.com/docs/api/#attributes). Maybe this should be a type instead of a type hole?

PR #13, opened 2017-06-23
Do retries on error

cyx:
We're currently getting intermittent errors. The weird part is they look like these:

```
<html> <head><title>400 Bad Request</title></head> <body bgcolor="white"> <center><h1>400 Bad Request</h1></center> <hr><center>awselb/2.0</center> </body> </html> 
```

It definitely seems non-retriable, but given the fact that we only get this intermittently suggests to me they're some weird shenanigans that can be retried.

This PR is a quick stab on retries. Tests to come if we even then it's worth it.

### DIFF
--- a/go-kit/metrics/provider/librato/librato.go
+++ b/go-kit/metrics/provider/librato/librato.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"net/url"
 	"sync"
@@ -32,6 +33,10 @@ var (
 type Error struct {
 	code                             int
 	body, rateLimitAgg, rateLimitStd string
+
+	// Used to debug things on occasion if inspection of the original
+	// request is necessary.
+	Request *http.Request
 }
 
 // Code returned by librato
@@ -65,6 +70,7 @@ type Provider struct {
 	source, prefix, percentilePrefix string
 	resetCounters, ssa               bool
 	numRetries                       int
+	requestDebugging                 bool
 
 	once sync.Once
 	done chan struct{}
@@ -91,6 +97,14 @@ type OptionFunc func(*Provider)
 func WithRetries(n int) OptionFunc {
 	return func(p *Provider) {
 		p.numRetries = n
+	}
+}
+
+// WithRequestDebugging enables request debugging and exposes the original
+// request in the Error.
+func WithRequestDebugging() OptionFunc {
+	return func(p *Provider) {
+		p.requestDebugging = true
 	}
 }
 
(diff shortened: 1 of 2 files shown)


freeformz:
I wonder if this is related to batching being missing (I promised it, but have been busy)? Can you ask librato about that possibility?

cyx:
Yeah currently asking. The elb error though is suspect.

cyx:
(e.g. in the past when we get a 400 from librato, it would be a JSON response, not an HTML one. But still waiting for a response from them).

freeformz:
This 400 is from the ELB. We may need to dump the complete request including headers and the like to debug this. :-(

niklibrato:
Yes, if you can capture the request including headers we can help debug the issue

go-kit/metrics/provider/librato/librato.go
```diff
@@ -285,19 +322,40 @@ func (p *Provider) report(u *url.URL, interval time.Duration) error {
 	if err := e.Encode(r); err != nil {
 		return err
 	}
-	resp, err := http.Post(u.String(), "application/json", &buf)
+
+	var rawRequest []byte
+	if p.requestDebugging {
+		rawRequest = buf.Bytes()
```
freeformz: You need to make a copy of the returned slice here as per the docs for Bytes() because http.DefaultClient.Do() will Read() from the buffer.

go-kit/metrics/provider/librato/librato.go
```diff
@@ -238,6 +262,19 @@ type attributes struct {
...
 }
 
+// reportWithRetry the metrics to the url, every interval, with max retries.
+func (p *Provider) reportWithRetry(u *url.URL, interval time.Duration) error {
+	var err error
+
+	for i := p.numRetries; i > 0; i-- {
+		if err = p.report(u, interval); err == nil {
+			return nil
+		}
```
freeformz: If the err != nil we should probably call p.errorHandler as well and possibly modify the error to indicate if it's final or will be retried. Otherwise we only get a chance to notify the user of the library of the last error.

go-kit/metrics/provider/librato/librato.go
```diff
@@ -238,6 +256,17 @@ type attributes struct {
...
 }
 
+// reportWithRetry the metrics to the url, every interval, with max retries.
+func (p *Provider) reportWithRetry(u *url.URL, interval time.Duration) {
+	for i := p.numRetries; i > 0; i-- {
+		err := p.report(u, interval)
+		if err == nil {
+			return
+		}
+		p.errorHandler(err)
```
freeformz: `errorHandler` may be `nil` and and IMO the error passes should have a way to indicate if it's temporary() or not.

PR #15, opened 2017-06-30
librato: Add batching

freeformz:
Batching is recommended behavior by Librato.

This also makes the requests to librato not drop metrics on retries, unless the
requests themselves can't be delivered. Currently a retry re-samples the metrics and looses any data previously sampled.


freeformz:
I realized today that there is likely a bug in this.

cyx:
Cool. Haven't deployed it, good to catch it early :)

freeformz:
@cyx This is probably ready to test out now.

freeformz:
@cyx Note, there are changes to the Error type, which now returns the dumped request.

PR #16, opened 2017-07-01
s3env: quick stab

cyx:
## Context

We're running out of config var space in our services and API doesn't seem to have a near term plan to increase this limit.

The goal of this tool is to provide a simple shim that will retrieve config from S3, and make the existing tools operate as if the env vars are still there.

## TODOS

- [ ] Need tests
- [ ] Think about multi string things (currently leaning on just doing envdecode style `;` separated things)
- [ ] Not everything in main? (Maybe yagni/meh. @freeformz have opinions?)


cmd/s3env/main.go
```diff
@@ -0,0 +1,88 @@
...
+		os.Exit(1)
+	}
+
+	exe, err := exec.LookPath(cmd)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "fatal: %s: %s\n", os.Args[1], err)
+		os.Exit(1)
+	}
+
+	if err = syscall.Exec(exe, nil, toEnviron(env)); err != nil {
```
jkakar: Does the value returned by `toEnviron` need to also include the environment variables that don't come from S3?

cyx: Good catch! I believe so :)

PR #19, opened 2017-08-11
hmetrics: Simplify error reporting and other cleanups

freeformz:
This simplifies the error reporting with an error handler letting
the user determine how to best handle the error.

Returning an error from the error handler abort metrics reporting
as well, giving the user an escape valve.

Additionally make the sharing of stats with ReadMemStats explicit.

### DIFF
--- a/hmetrics/init.go
+++ b/hmetrics/init.go
@@ -63,7 +63,8 @@ func Report(ctx context.Context, ef ErrHandler) error {
 	return nil
 }
 
-// The only thing that should come after and exit() is a return
+// The only thing that should come after an exit() is a return.
+// Best to use in a function that can defer it.
 func exit() {
 	mu.Lock()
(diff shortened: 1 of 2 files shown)


hmetrics/init.go
```diff
@@ -55,31 +55,45 @@ func Report(ctx context.Context, pfer printfer) error {
...
 	defer t.Stop()
 
 	for {
 		select {
 		case <-t.C:
 		case <-ctx.Done():
-			mu.Lock()
-			defer mu.Unlock()
-			started = false
+			exit()
```
bernerdschaefer: Can this be a single `defer exit()` -- or even inlined again `defer func() { mu.Lock() ...}` -- outside the loop?

freeformz: good call. Done.

PR #20, opened 2017-08-11
hmetrics: Other metrics I think we should report

freeformz:
Also, I'm not sure calculating the derivative for pause times is the right thing to do here.
We don't do the same thing with the number of GC runs.


freeformz:
@apg How do metaas counters work? We're calculating the derivative for the pause time, but not the numbers of GC runs. So I'm not sure which one is correct?

apg:
Counters are reported as derivatives. So, yeah, there's a bug if it's reporting an absolute count.

PR #24, opened 2017-10-04
A bunch of cleanups; see comments

freeformz:
Document everything more, including usage of all exported bits.

Reduce the use of global variables down to the started flag and it's mutex.

Report is now synchronous, pushing that off to the caller if they directly use
Report. onload is unaffected. Cleaned up examples to reflect this.

Fixed the package name declaration in onload (was onloads by accident).

Renamed a few functions and variables for clarity.

### DIFF
--- a/hmetrics/hmetrics.go
+++ b/hmetrics/hmetrics.go
@@ -1,12 +1,19 @@
+/* Copyright (c) 2018 Salesforce
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root  or https://opensource.org/licenses/BSD-3-Clause
+ */
 package hmetrics
 
 import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"runtime"
 	"sync"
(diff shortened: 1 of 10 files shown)


hmetrics/hmetrics.go
```diff
@@ -44,61 +50,84 @@ func (e HerokuMetricsURLUnset) Fatal() bool {
...
+//
+// Report is safe for concurrent usage, but calling it more than once w/o
+// canceling the context returns an AlreadyStarted error. This is to ensure that
+// metrics aren't duplicated. Report can be called again to restart reporting
+// after the context is canceled.
 func Report(ctx context.Context, ef ErrHandler) error {
 	mu.Lock()
-	defer mu.Unlock()
 	if started {
+		mu.Unlock()
```
apg: Can we just use `sync.Once` for all this?

hmetrics/hmetrics.go
```diff
@@ -34,6 +38,8 @@ func (as AlreadyStarted) Fatal() bool {
 	return false
 }
 
+// HerokuMetricsURLUnset represents the Error condition when the
+// HEROKU_METRICS_URL environment variables is unset or an empty string.
 type HerokuMetricsURLUnset struct{}
```
murphybytes: variables -> variable

hmetrics/hmetrics.go
```diff
@@ -44,61 +50,84 @@ func (e HerokuMetricsURLUnset) Fatal() bool {
 	return true
 }
 
+// ErrHandler funcations are used to provide custom processing/handling of
+// errors encountered during the collection or reporting of metrics to Heroku.
+type ErrHandler func(err error) error
+
```
murphybytes: funcations -> functions

hmetrics/hmetrics.go
```diff
@@ -44,61 +50,84 @@ func (e HerokuMetricsURLUnset) Fatal() bool {
...
 
-// ErrHandler receives any errors encountered during collection or reporting of metrics to Heroku. Processing of metrics
-// continues if the ErrHandler returns nil, but aborts if the ErrHandler itself returns an error.
-type ErrHandler func(err error) error
-
+// Report go metrics to Heroku until the context is canceled.
+//
+// Only one call to the ErrHandler will happen at a time. Metrics can be dropped
+// or delayed if a call to ErrHandler takes longer than the reprting interval.
+// Processing of metrics continues if the ErrHandler returns nil, but aborts if
```
murphybytes: reprting -> reporting

hmetrics/hmetrics.go
```diff
@@ -44,61 +50,84 @@ func (e HerokuMetricsURLUnset) Fatal() bool {
...
+// the ErrHandler itself returns an error. It is safe to pass a nil ErrHandler.
+//
+// Report is safe for concurrent usage, but calling it more than once w/o
+// canceling the context returns an AlreadyStarted error. This is to ensure that
+// metrics aren't duplicated. Report can be called again to restart reporting
+// after the context is canceled.
 func Report(ctx context.Context, ef ErrHandler) error {
 	mu.Lock()
-	defer mu.Unlock()
 	if started {
```
murphybytes: I would break the code before and after report into two functions to make it clearer what's happening here.  As it is it's kind of confusing, maybe the following?  Also, I'm thinking that the public DefaultEndpoint is going to be a problem for concurrent use.
```
func Report(ctx context.Context, ef ErrHandler) error 
       if endpoint, err := alreadyStarted(); err != nil {
            return err 
       }
	report(ctx, &http.Client{Timeout: 20 * time.Second}, endpoint, errorHandler(ef))
        setStarted() 
	return nil
}

func alreadyStarted() string, error {
      mu.Lock() 
      defer  func() {
          started = localStarted
          mu.Unlock() 
     }() 
     var localStarted bool
      
	if started {
		return "", AlreadyStarted{}
	}
        
	if DefaultEndpoint == "" {
		return "" HerokuMetricsURLUnset{}
	}
	localStarted = true
        endpoint := DefaultEndpoint
        return endpoint,  nil 
}
```

hmetrics/hmetrics.go
freeformz: No. Because it could be restarted.

hmetrics/hmetrics.go
freeformz: @murphybytes I did something similar, but different. Please take a look and if you are okay, approve. Thx.

philpennock:
Folks, there are people actually using this software, changing an async function to a synchronous function without also changing the name is downright hostile.  Someone on a project I worked on made the parameter change to comply but missed the change in behavior and need to revamp calling.

Please consider carefully whether or not to rename a function which changes behavior so critically.

ydnar:
Yes. This change caused hard downtime affecting our customers.

freeformz:
@ydnar @philpennock Ugh, sorry about that. I always wanted this to be sync (a sync API is easier to make async than vice versa; generally easier to reason about, etc.) and since the feature isn't GA and people generally vendor I didn't consider the breaking change too heavily. Sorry. :-( We'll do better.

PR #25, opened 2017-11-15
hmetrics/onload: spew metrics errors when HMETRICS_VERBOSE is set truthy

Xe:
Currently package `hmetrics` in the `onload` configuration ignores all errors instead of logging them. This can make debugging of metrics collection errors very difficult and time-consuming. This patch offers a middle path between every error always being logged by default and no errors being logged. When the user sets `HMETRICS_VERBOSE=1` in the environment, all errors will be logged when they happen.

## Stories this patch fulfills
```
As a user of package hmetrics in customer code
In order to better debug metrics collection errors
Given the user set HMETRICS_VERBOSE to 1
When there is a reporting error
Then the reporting error is printed to standard error
```
```
As a user of package hmetrics in customer code
In order to better debug metrics collection errors
Given the user set HMETRICS_VERBOSE to 0
When there is a reporting error
Then the reporting error is not printed to standard error
```

### DIFF
--- a/hmetrics/onload/init.go
+++ b/hmetrics/onload/init.go
@@ -11,10 +11,10 @@ retrying reporting, backing off in 10 second increments.
 Use this package when you don't care about stopping reporting, specifying the
 endpoint, or being notified of any reporting errors.
 
-This package additionally will report all metrics reporting errors using package
-log to the standard error file descriptor if you set the environment variable
-`HMETRICS_VERBOSE` to `1` or another true-like value as defined by
-https://godoc.org/strconv#ParseBool.
+This package will log all reporting errors via the stdlib log package if the
+environment variable HMETRICS_VERBOSE is set to "1", or any other true-like
+value as defined by strconv#ParseBool (see $ godoc strconv for more
+information).
 
 usage:
 
@@ -43,29 +43,27 @@ const (
 )
 
 func init() {
+	var logger hmetrics.ErrHandler = func(_ error) error { return nil }
+
+	val := os.Getenv("HMETRICS_VERBOSE")
+	should, err := strconv.ParseBool(val)
+	if err == nil && should {
+		logger = func(err error) error {
+			log.Printf("[hmetrics] error: %v", err)
+			return nil
+		}
+	}
+
 	go func() {
 		var backoff int64
 		for backoff = 1; ; backoff++ {
 			start := time.Now()
 
-			var logger hmetrics.ErrHandler
-
-			val := os.Getenv("HMETRICS_VERBOSE")
-			should, err := strconv.ParseBool(val)
-			if err == nil && should {
-				logger = func(err error) error {
-					log.Printf("[hmetrics] error: %v", err)
-					return nil
-				}
-			}
-
 			err = hmetrics.Report(context.Background(), hmetrics.DefaultEndpoint, logger)
 			if time.Since(start) > 5*time.Minute {
 				backoff = 1
 			}
-			if logger != nil {
-				logger(err)
-			}
+			_ = logger(err)
 
 			time.Sleep(time.Duration(backoff*interval) * time.Second)
 		}
--- a/hmetrics/onload/init.go
+++ b/hmetrics/onload/init.go
@@ -63,7 +63,7 @@ func init() {
 			if time.Since(start) > 5*time.Minute {
 				backoff = 1
 			}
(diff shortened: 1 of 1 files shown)


freeformz:
What is the reason behind this? I don't like enabling logging in libraries by default and the library as a whole provides a mechanism to do this when it's deemed necessary.

justincampbell:
This is identical to the code I ended up adding in our application. I'd love to at least see this in the README or added to the [basic example](https://github.com/heroku/x/blob/master/hmetrics/example/basic/main.go#L13) (which is what I ended up starting from).

Xe:
As of today I have rebased this commit on master and made the changes I want in `github.com/heroku/x/hmetrics/onload-log`. This package will function identically to the other metrics onload package except all errors will be logged with [package log](https://golang.org/pkg/log).

@freeformz @murphybytes can you please re-review?

hmetrics/onload-log/init.go
```diff
@@ -0,0 +1,58 @@
+/* Copyright (c) 2018 Salesforce
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root  or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/*
+Package onload automatically starts hmetrics reporting, ignoring errors and
```
freeformz: s/onload/onload-log/
s/ignoring errors/logging errors using the stdlib log package/

hmetrics/onload-log/init.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/heroku/x/hmetrics"
+)
+
+const (
+	interval = 10
+)
+
+// logError forwards the metrics submission error to package log.
+func logError(err error) error {
+	log.Printf("[hmetrics] error when submitting metrics: %v", err)
```
freeformz: s/when//

hmetrics/onload/init.go
```diff
@@ -39,10 +47,26 @@ func init() {
...
 		for backoff = 1; ; backoff++ {
 			start := time.Now()
-			hmetrics.Report(context.Background(), hmetrics.DefaultEndpoint, nil)
+
+			var logger hmetrics.ErrHandler
+
+			val := os.Getenv("HMETRICS_VERBOSE")
+			should, err := strconv.ParseBool(val)
+			if err == nil && should {
+				logger = func(err error) error {
```
freeformz: similarly to @apg above recommendation this could be constructed above.

hmetrics/onload/init.go
```diff
@@ -39,10 +47,26 @@ func init() {
...
+					return nil
+				}
+			}
+
+			err = hmetrics.Report(context.Background(), hmetrics.DefaultEndpoint, logger)
 			if time.Since(start) > 5*time.Minute {
 				backoff = 1
 			}
+			if logger != nil {
+				logger(err)
```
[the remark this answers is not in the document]
freeformz: We could construct a noop logger and skip the check(s).

hmetrics/onload/init.go
```diff
@@ -43,29 +43,27 @@ const (
...
-			}
-
 			err = hmetrics.Report(context.Background(), hmetrics.DefaultEndpoint, logger)
 			if time.Since(start) > 5*time.Minute {
 				backoff = 1
 			}
-			if logger != nil {
-				logger(err)
-			}
+			_ = logger(err)
```
freeformz: Why not `logger(err)` ?

hmetrics/onload/init.go
```diff
@@ -35,14 +43,28 @@ const (
...
+		}
+	}
+
 	go func() {
 		var backoff int64
 		for backoff = 1; ; backoff++ {
 			start := time.Now()
-			hmetrics.Report(context.Background(), hmetrics.DefaultEndpoint, nil)
+
+			err = hmetrics.Report(context.Background(), hmetrics.DefaultEndpoint, logger)
```
freeformz: s/`err = `// if we're not going to do anything with it.

hmetrics/onload/init.go
Xe: fixed in https://github.com/heroku/x/pull/25/commits/6ed42b46be7bbf1e35ee04b29d6f2e86aa04d04e

PR #28, opened 2018-03-08
Ensure all tests pass, set up travis

### LINKED ISSUE #27: Set up travis
People depend on this repo. We should set up automated CI.

## Checklist

- [ ] heroku/x#28

Xe:
Fixes #27

This PR ensures all code passes building and tests. This also includes vendored dependencies.


.travis.yml
```diff
@@ -1,8 +1,6 @@
...
-go: 1.8.3
+go: '1.10'
 sudo: false
 install: true
 script:
-- go get -u -v ./...
-- go test -v ./...
-- go test -v -race ./...
\ No newline at end of file
+- go test -v -race ./...
```
freeformz: Does travis cache anything? If so we may need to add `-count=1` to ensure that we don't have a test that skirts cache busting.

cache/cache.go
```diff
@@ -1,37 +0,0 @@
-// Package cache is a generic cache mechanism.
```
freeformz: The description doesn't mention anything about removing cache. What's the reason for removing it?

Xe: I plan to re-introduce this later, properly isolating it, making it context and more generically metrics and logging-aware.

PR #29, opened 2018-03-13
hcontext: Add documentation, examples, tests

Xe:
- [x] Document everything better
  - [x] functions
  - [x] types
  - [x] package commentary
  - [x] blocking vs nonblocking (if applicable)
- [x] Example usage
- [x] tests

### DIFF
--- a/hcontext/requestid.go
+++ b/hcontext/requestid.go
@@ -15,13 +15,14 @@ import (
 
 type idkey int
 
-var key idkey
+var ridKey idkey
 
 var headersToSearch = []string{
-	"Request-ID", "X-Request-ID",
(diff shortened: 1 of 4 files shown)


hcontext/requestid.go
```diff
@@ -15,30 +15,34 @@ import (
 
 type idkey int
 
-var key idkey
+var ridKey idkey
 
 var headersToSearch = []string{
-	"Request-ID", "X-Request-ID",
+	"Request-Id", "X-Request-Id",
```
amerine: Is this a breaking change?

Xe: Oh god, everything is awful. I'm just gonna specify both.

PR #35, opened 2018-05-31
Add a HyperLogLog based Unique Counter for high cardinality, unique sets

apg:
This adds a new type to our custom provider, UniqueCounter, which allows one to provide a metric type that counts the cardinality of a set of strings. The implementation for the Librato provider uses a HyperLogLog estimator, which reports as a gauge, and is capable of being reset (just like the standard counters provided).

I'm not married to the name `UniqueCounter` but, I wasn't clever enough to come up with a more suitable name that provided the same obvious meaning.


apg:
As an aside, I tried to do this without modifying the Librato provider, and that was a train wreck. The idea was to take a `*hll.Sketch` and a `metrics.Gauge` and keep the `Gauge` in sync with the Sketch's estimate. But, as it turns out `Sketch.Estimate` isn't exactly fast, so the overhead of this was something like 500x _slower_ than just updating the `Gauge`, or a regular `Counter`. Obviously, the gauge doesn't need to be updated every time, but coordinating it all, in the presence of a reporter that samples occasionally, just didn't seem ideal...

apg:
Ugh.. I guess, also, I've gotta vendor the library, which seems horrible....

go-kit/metrics/provider/librato/uniquecounter.go
```diff
@@ -0,0 +1,75 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root  or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+package librato
+
+import (
+	"sync"
+
+	hll "github.com/axiomhq/hyperloglog"
```
freeformz: Does this hyperloglog implementation do what we think it does? I read through the docs and it says that it is "An improved version of HyperLogLog for the count-distinct problem, approximating the number of distinct elements in a multiset using 20-50% less space than other usual HyperLogLog implementations."

Given that the cardinality of a multiset is different (sum of the multiplicities of the multiset's members) than the cardinality of a set (count of the elements in the set) I am confused.

go-kit/metrics/provider.go
```diff
@@ -22,5 +22,6 @@ type Provider interface {
 	NewCounter(name string) metrics.Counter
 	NewGauge(name string) metrics.Gauge
 	NewHistogram(name string, buckets int) metrics.Histogram
+	NewUniqueCounter(name string) UniqueCounter
```
freeformz: Since the addition of this method makes it deviate from the documented Provider please update the docs above.

go-kit/metrics/provider/librato/librato_test.go
```diff
@@ -532,3 +538,47 @@ func TestWithResetCounters(t *testing.T) {
 		})
 	}
 }
+
+func TestWithResetCountersUniqueCounters(t *testing.T) {
+	//	for _, reset := range []bool{true, false} {
+	for _, reset := range []bool{true} {
```
freeformz: ,false ?

go-kit/metrics/provider/librato/uniquecounter.go
```diff
@@ -0,0 +1,75 @@
...
+
+func NewHLLCounter(name string) *HLLCounter {
+	return &HLLCounter{
+		Name:    name,
+		counter: hll.New(),
+	}
+}
+
+// With returns a new UniqueCounter with the passed in label values merged
+// with the previous label values. The counter value is forked.
```
freeformz: Suggested edit: The counter's values are copied.

go-kit/metrics/provider/librato/uniquecounter.go
freeformz: Hmm. The demo seems to indicate it does.

go-kit/metrics/provider/librato/uniquecounter_test.go
```diff
@@ -0,0 +1,36 @@
+/* Copyright (c) 2018 Salesforce
```
freeformz: Re: filename ... This does not test the uniquecounter it tests `HLLCounter`.

freeformz:
vendor is there for our testing purposes and would be stripped by other tools that vendor this library.

go-kit/metrics/provider/librato/histogram.go
```diff
@@ -0,0 +1,145 @@
+/* Copyright (c) 2018 Salesforce
```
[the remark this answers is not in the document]
apg: Fair.

go-kit/metrics/uniquecounter.go
```diff
@@ -0,0 +1,14 @@
+/* Copyright (c) 2018 Salesforce
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root  or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+package metrics
+
+// UniqueCounter describes a metric that reports a count of the number of
```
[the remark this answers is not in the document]
apg: I like the idea of Set, but `Estimate` on a Set doesn't really make sense. I previously used `CardinalityCounter` in another context, and while I liked that, it was confusing to some people who didn't understand the concept of cardinality. Perhaps that's the best choice though...

go-kit/metrics/provider/librato/uniquecounter.go
apg: Given the way that HLL works, I don't see how it actually can estimate the "sum of the multiplicities" as it uses a set of hashes that are deterministic. It's been a while since I've read any of the literature around this, though, so maybe I'm wrong.

go-kit/metrics/provider/librato/librato_test.go
apg: Ah. Yes. I was doing some debugging and forgot to back this one out.

go-kit/metrics/provider/librato/uniquecounter_test.go
apg: Fair enough.

go-kit/metrics/provider/librato/uniquecounter.go
```diff
@@ -0,0 +1,75 @@
...
+	defer c.mu.Unlock()
+	c.counter.Insert(x)
+}
+
+// Estimate provides the counter's current estimate of cardinality.
+func (c *HLLCounter) Estimate() uint64 {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	return c.counter.Estimate()
```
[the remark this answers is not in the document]
apg: That doesn't pan out. My comment on suggesting Estimate was expensive was mostly around keeping the Gauge in sync, e.g., calling `Estimate()` for every `Insert` instead of doing a atomic CAS swap, like the Gauge, or an atomic increment, like the Counter.

go-kit/metrics/provider/librato/uniquecounter.go
```diff
@@ -0,0 +1,75 @@
...
+	hll "github.com/axiomhq/hyperloglog"
+	xmetrics "github.com/heroku/x/go-kit/metrics"
+)
+
+var (
+	_ xmetrics.UniqueCounter = &HLLCounter{}
+)
+
+// HLLCounter provides a wrapper around a HyperLogLog probabalistic
+// counter, capable of being reported to Librato.
```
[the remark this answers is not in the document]
apg: You could say the same thing about the histogram implementation...

apg:
@freeformz OK, I guess a rereview would work here.

go-kit/metrics/provider.go
```diff
@@ -14,13 +16,17 @@ import (
...
 
-// Provider represents all the kinds of metrics a provider must expose. This is
-// here for 2 reasons: (1) go-kit/metrics/provider imports all the providers in
-// the world supported by go-kit cluttering up your vendor folder; and (2)
-// provider.Provider (hmmmmm stutter)!
+// Provider represents the different types of metrics that a provider
+// can expose. We duplicate the definition from go-kit for 2 reasons:
+//
+//  1. A little copying never hurt anyone (and in copying, we avoid the
+//       need to import and vendor all of go-kit's supported providers
```
danp: Minor: this indentation is a bit off from the indentation for the second line of 2.

apg: no sacrifice too great!

go-kit/metrics/provider/librato/librato_test.go
```diff
@@ -532,3 +538,46 @@ func TestWithResetCounters(t *testing.T) {
 		})
 	}
 }
+
+func TestWithResetCountersUniqueCounters(t *testing.T) {
```
benburkert: `s/Unique/Cardinatilty/`

go-kit/metrics/hllcounter.go
```diff
@@ -0,0 +1,76 @@
...
+// HLLCounter provides a wrapper around a HyperLogLog probabalistic
+// counter, capable of being reported to Librato.
+type HLLCounter struct {
+	Name    string
+	lvs     []string
+	mu      sync.RWMutex
+	counter *hll.Sketch
+}
+
+func NewHLLCounter(name string) *HLLCounter {
```
benburkert: needs docs.

go-kit/metrics/hllcounter.go
```diff
@@ -0,0 +1,76 @@
...
+var (
+	_ CardinalityCounter = &HLLCounter{}
+)
+
+// HLLCounter provides a wrapper around a HyperLogLog probabalistic
+// counter, capable of being reported to Librato.
+type HLLCounter struct {
+	Name    string
+	lvs     []string
+	mu      sync.RWMutex
```
benburkert: Assuming `mu` will mostly be locked for `Insert()` calls, a plain old `sync.Mutex` might be more performant.

apg:
OK. I've addressed @freeformz's major concerns and gotten 2 other reviews. I'm going to merge this, and if he still has additional concerns later, we can always open another PR. As it stands, I'd like to incorporate this, though, into my own repo.

PR #37, opened 2018-06-14
tags: phase 1 support

cyx:
## Context

As part of doing our HTTP availability work in @heroku/runtime-tbd we've begun to explore the possibility of using a librato account with tags.

In order to start supporting tags, we'll want to change the semantics of `With(...)` to first of all:

1. Work
2. Register itself to something

## Phase 1

Phase 1 is to just splat out the label values into the name using a convention, e.g. given:

```
prefix: heroku
name: http_request
tags: { region: us, app: myapp, space: myspace }
```

It computes to:

```
heroku.http_request.region:us.app:myapp.space:myspace
```

## Next steps

The next step is to separate out the reporter into 2 concrete interfaces with the same interface. A v1 and a v2 which will be switched to v2 automatically when you do `WithTags`.

The interface is drastically different that it seems to warrant this, specifically:

1. It seems that it wants to post to `/v1/measurements` instead for tags
2. `measure_time` is no longer a global thing in the payload, but per measurement
3. `sum_squares` is no longer a thing, but seems to now require `stddev`

There's a bunch more details which will be fleshed out in a follow up PR.

## Unknown / out of scope for now

`HLLCounter` seems to be significantly different in terms of implementation so I've left it out for now.

### DIFF
--- a/go-kit/metrics/provider/librato/librato.go
+++ b/go-kit/metrics/provider/librato/librato.go
@@ -242,16 +242,6 @@ func keyName(name string, labelValues ...string) string {
 
 // metricName returns the name we'll use for the metric depending on
 // whether we're using tags or not.
-//
-// GIVEN
-//   name = requests
-//   tags = { region = us, app = myapp }
-//
-// TAGS ENABLED:
-//   requests
-//
-// TAGS DISABLED:
-//   requests.region:us.app:myapp
 func (p *Provider) metricName(name string, labelValues ...string) string {
 	if p.tagsEnabled {
 		return name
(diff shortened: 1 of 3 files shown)


go-kit/metrics/provider/librato/librato.go
```diff
@@ -211,35 +225,119 @@ func prefixName(prefix, name string) string {
...
+	for i := 0; i < l; i += 2 {
+		parts = append(parts, labelValues[i]+":"+labelValues[i+1])
+	}
+	return name + "." + strings.Join(parts, ".")
+}
+
+// metricName returns the name we'll use for the metric depending on
+// whether we're using tags or not.
+//
+// GIVEN
```
benburkert: looks like this behavior is captured in the `TestProviderMetricName` test so i think this comment block is unnecessary.

PR #38, opened 2018-06-14
Batcher: extract out so we can build a different batcher

cyx:
- Move histogram.measures out since it's hard coded to pre-tags
- Adds new test for tag based stuff

### DIFF
--- a/go-kit/metrics/provider/librato/batcher.go
+++ b/go-kit/metrics/provider/librato/batcher.go
@@ -1,0 +1,424 @@
+package librato
+
+import (
+	"bytes"
+	"encoding/json"
+	"math"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+const (
+	oldBatcherPath = "/v1/metrics"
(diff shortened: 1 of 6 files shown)


go-kit/metrics/provider/librato/batcher.go
```diff
@@ -0,0 +1,424 @@
...
+	p *Provider
+}
+
+// Batch will batch up all the metrics into []*http.Requests using the old style API.
+func (b *oldBatcher) Batch(u *url.URL, interval time.Duration) ([]*http.Request, error) {
+	// Calculate the sample time.
+	st := time.Now().Truncate(interval).Unix()
+
+	// Sample the metrics.
+	gauges := b.sample(int(interval.Seconds()))
```
benburkert: given my earlier comment, maybe this can turn into:

``` go
measurements := b.sample(int(interval.Seconds()))
gauges = make([]gauge, 0, len(measurements))
for _, m := range measurements {
  gauges = append(gauges, m.Gauge())
}
```

PR #39, opened 2018-06-15
metrics: cleanup and add more tests

cyx:
Previously, `histogramMeasures` wasn't removed properly. That
has been removed, but also fixed by adding the tags / attributes
in all relevant places -- and also added more table tests for
the scenario of tags vs no tags.

### DIFF
--- a/go-kit/metrics/provider/librato/batcher.go
+++ b/go-kit/metrics/provider/librato/batcher.go
@@ -99,41 +99,6 @@ type gauge struct {
 	SumSq  float64 `json:"sum_squares"`
 }
 
-// the json marshalers for the histograms 4 different gauges
-func (b *oldBatcher) histogramMeasures(h *Histogram, period int) []gauge {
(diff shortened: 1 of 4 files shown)


go-kit/metrics/provider/librato/batcher.go
```diff
@@ -196,9 +161,9 @@ type measurement struct {
 	Attributes map[string]interface{} `json:"attributes,omitempty"`
 	Tags       map[string]string      `json:"tags"`
 
+	Count  int64   `json:"count"`
```
apg: This rearrangement seems unnecessary, but I'm going to guess that it's sole purpose is to make the JSON marshaling align correctly?

cyx: Yep :)

cyx: Also it's an aesthetic in terms of int64 being separated from the float64s.

PR #41, opened 2018-06-19
go-kit/metrics/provider/librato: combine batcher types, add default tags

benburkert:
Combine the `batcher` types into a single struct type with a `batchMetrics` for tagless and `batchMeasurements` for tagged metrics.

Add tag arguments to the `librato.WithTags` opt func. Tags specified in the `WithTags` call are included in all metrics created by the provider.

### DIFF
--- a/go-kit/metrics/provider/librato/librato_test.go
+++ b/go-kit/metrics/provider/librato/librato_test.go
@@ -341,7 +341,7 @@ func TestLibratoSingleReportWithLabelValuesOnTagBasedAccount(t *testing.T) {
 		t.Fatal("unexpected error reporting metrics", err)
 	}
 
-	p := New(u, doesntmatter, WithTags(), WithSource("test.source"), WithErrorHandler(errs))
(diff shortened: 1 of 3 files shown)


go-kit/metrics/provider/librato/librato_test.go
```diff
@@ -341,7 +341,7 @@ func TestLibratoSingleReportWithLabelValuesOnTagBasedAccount(t *testing.T) {
 		t.Fatal("unexpected error reporting metrics", err)
 	}
 
-	p := New(u, doesntmatter, WithTags(), WithSource("test.source"), WithErrorHandler(errs))
+	p := New(u, doesntmatter, WithTags("app", "myapp"), WithSource("test.source"), WithErrorHandler(errs))
```
apg: It'd be great to test that the tags show up?

cyx: +1

PR #42, opened 2018-06-22
 go-kit/metrics/provider/librato: fix tagged histogram metric name, only sample used metrics

benburkert:
Ignore unused (unobserved) metrics, which happens when an observer is only used to create tagged observers via `.With` method calls. Remove the `tagsEnabled` field from `batcher`, which was breaking tagging as it never was set. Use the `metricName()` helper to construct `Gauge` and `Counter` metric names.

### DIFF
--- a/go-kit/metrics/provider/librato/librato.go
+++ b/go-kit/metrics/provider/librato/librato.go
@@ -452,6 +452,14 @@ func (h *Histogram) SumSq() float64 {
 type Counter struct {
 	*generic.Counter
 	p *Provider
+
+	used bool
+}
+
+// Add implements Counter.
+func (c *Counter) Add(delta float64) {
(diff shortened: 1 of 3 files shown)


go-kit/metrics/provider/librato/librato.go
```diff
@@ -452,6 +452,14 @@ func (h *Histogram) SumSq() float64 {
 type Counter struct {
 	*generic.Counter
 	p *Provider
+
+	used bool
```
apg: A comment on this variable would be welcome. It's only because I have more context from speaking with y'all that I understand it's purpose (and why it doesn't get reset).

benburkert: :+1: Done.

PR #43, opened 2018-06-23
BUGFIX: s3env os.environ dropping things

cyx:


cmd/s3env/run.go
```diff
@@ -51,7 +51,7 @@ therefore all signals are properly delegated.`,
 // merge takes a map of envs and a slice and combines them into one slice.
 // e.g. given "A" => 1, and []{"B=2"}, you get {"A=1", "B=2"}
 func merge(env map[string]string, environ []string) []string {
-	result := make([]string, len(env), len(env)+len(environ))
+	result := make([]string, len(environ), len(env)+len(environ))
```
bernerdschaefer: @cyx Did you consider making this an empty list and appending both the host environ and s3env map? There doesn't seem to be any performance needs to justify this method, and trying to explicitly size this was the direct cause of the previous bug.

PR #46, opened 2018-07-12
Librato: add internal metrics

cyx:
## Context

We want internal metrics to be reported. For now, these are limited to:

1. The rate limits we're seeing
2. The total measurements sampled during reporting.

These measurements will allow us to monitor the provider itself (e.g. are we sending too many measurements) and potentially build alerts around them.

Connects to https://github.com/heroku/cedar/issues/2479

### DIFF
--- a/go-kit/metrics/provider/librato/librato.go
+++ b/go-kit/metrics/provider/librato/librato.go
@@ -203,9 +203,9 @@ func New(URL *url.URL, interval time.Duration, opts ...OptionFunc) metrics.Provi
 		}
 	}
 
-	p.measurements = p.NewGauge("internal.measurements")
-	p.ratelimitAgg = p.NewGauge("internal.ratelimit-aggregate")
-	p.ratelimitStd = p.NewGauge("internal.ratelimit-standard")
+	p.measurements = p.NewGauge("go-kit.measurements")
+	p.ratelimitAgg = p.NewGauge("go-kit.ratelimit-aggregate")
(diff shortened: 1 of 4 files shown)


go-kit/metrics/provider/librato/librato.go
```diff
@@ -199,6 +203,10 @@ func New(URL *url.URL, interval time.Duration, opts ...OptionFunc) metrics.Provi
 		}
 	}
 
+	p.measurements = p.NewGauge("internal.measurements")
```
bernerdschaefer: What will the prefix and source/tags end up as for these?

cyx: Whatever you use in the project. So if you're on probedom, `probedom.internal.measurements`. if on dyno-funnel`, `cedar.dyno-funnel.internal.measurements`. Source / etc are applied as normal.

benburkert: `internal` is a bit vague IMO, something like `go-kit` would be better.

go-kit/metrics/provider/librato/sample.go
```diff
@@ -173,6 +173,10 @@ func (p *Provider) sample(period int) []measurement {
 		})
 	}
 
+	if p.measurements != nil {
```
benburkert: should this check `len(p.measurements) > 0`?

benburkert: nevermind, `p.measurements` is not a slice.

PR #47, opened 2018-07-13
Remove used concept for now

cyx:
This can cause pages to not self resolve if we don't emit the 0s.


bernerdschaefer:
Why was `used` originally introduced, and has anything changed to no longer make that an issue?

cyx:
> Why was used originally introduced, and has anything changed to no longer make that an issue?

It's mostly in relation to tags. We didn't want to see metrics that had empty tags in them (since the base metrics typically don't get added to, not without a `With(...)`.

The dogwood router relies on this, so it can keep it's vendor version for now.

PR #50, opened 2018-07-19
histogram: replace the underling library

cyx:
## Context

The previous library we were using suffered from some issues, allocation was one of them, and would mostly causes a lot of garbage.

Here's a comparison of the two libraries with the typical options:

```
BenchmarkTDigestCompression-8   	 5000000	       254 ns/op	       3 B/op	       0 allocs/op
BenchmarkVividCortex-8          	 2000000	       918 ns/op	    3470 B/op	       5 allocs/op
PASS
ok  	_/Users/cdavid/histogram	4.334s
```

[see the micro benchmark here](https://gist.github.com/cyx/318dc5853aa6daf7f862b749489d9147)

## Description of changes

- Replaces `gohistogram`.
- Fixes test

Note that even with a small sample size, you can see that `tdigest` offers better granularity (e.g. the previous `min` / `max` values actually improve.


go-kit/metrics/provider/librato/librato.go
```diff
@@ -381,7 +381,7 @@ type Histogram struct {
 	mu sync.RWMutex
 	// I would prefer to use hdrhistogram, but that's incompatible with the
```
apg: Can probably kills this comment.

go-kit/metrics/provider/librato/librato.go
```diff
@@ -416,7 +416,7 @@ func (h *Histogram) With(labelValues ...string) kmetrics.Histogram {
 
 func (h *Histogram) reset() {
 	// Not happy with this, but the existing histogram doesn't have a Reset.
-	h.h = gohistogram.NewHistogram(h.buckets)
+	h.h, _ = tdigest.New() // errors only happen if you pass in wrong options.
```
apg: Kill the _above_ comment (not happy...)

Make this comment: "`err` is only non-nil when passing options."

apg:
yippeee!

PR #53, opened 2018-11-02
librato: include response headers in error

bernerdschaefer:
With debug enabled, reported errors from the Librato package include the
full request sent to Librato, including headers. The errors also include
the response body -- but not the response headers.

The response headers are helpful for debugging failed requests, as well,
like for narrowing down the source of a rejected request.

This updates the `Body` of error responses to include the (scrubbed)
headers of the response as well as the text of the body.


sowjumn:
Looks good to me.

PR #54, opened 2019-05-03
use go.mod; add testing testenv

cyx:
## Context

We wanna be able to import this easily to a project without causing conflicts. This will be place to put heroku/runtime generic things that we've been using and so we want this to import correctly in a place like heroku/lyra-labs.

Adding `testing/testenv` for starters which is used heavily for testing in heroku/runtime.


jingweno:
LGTM

PR #56, opened 2019-05-06
Kit metrics from runtime

cyx:
## Context

We want to start chunking out libraries from runtime so we can start re-using these. Primarily, the focus is on metrics, cmd/internal things, and grpc bits.


cyx:
@jingweno just had to throw a mutex on that test histogram.

PR #59, opened 2019-05-10
Add redigo.Pool support for multiple passes.

apghero:
This modifies the interface for NewPool to optionally accept a list
of other passwords that might be correct, and iterates over them
caching the one that actually works, on first Dial.

Based on one of the options, discussed externally, for solving the problem of redis cred rolls.


hredis/redigo/redigo.go
```diff
@@ -42,15 +44,61 @@ func WaitForAvailability(url string, d time.Duration, f WaitFunc) (bool, error)
...
+
+				for _, pass := range passesToTry {
+					if _, err := c.Do("AUTH", pass); err == nil {
+						password = pass
+						return
+					}
+				}
+			})
+
+			// This is necessary since
```
apghero: Since what?

hredis/redigo/redigo.go
```diff
@@ -42,15 +44,61 @@ func WaitForAvailability(url string, d time.Duration, f WaitFunc) (bool, error)
...
+	// with which we can try all of the possible passwords.
+	u.User = url.UserPassword("", "")
+	rawURL = u.String()
+
 	return &redis.Pool{
 		MaxIdle:     3,
 		IdleTimeout: 240 * time.Second,
 		Dial: func() (redis.Conn, error) {
-			return redis.DialURL(url)
+			c, err := redisDialURL(rawURL)
```
danp: When using redis.DialURL in running code won't this be too late to try other passwords? If I'm reading right, DialURL in turn calls Dial which [will then try and AUTH with the given password](https://github.com/gomodule/redigo/blob/39e2c31b7ca38b521ceb836620a269e62c895dc9/redis/conn.go#L229-L234), erroring if it doesn't work.

apghero: Exactly why the password is cleared from the URL first. :) IF the password is blank, it won't issue `"AUTH"`

danp: Yargh, sorry. It's Friday! 🎉

hredis/redigo/redigo.go
```diff
@@ -42,15 +44,61 @@ func WaitForAvailability(url string, d time.Duration, f WaitFunc) (bool, error)
...
+			if err != nil {
+				return nil, err
+			}
+
+			findPassword.Do(func() {
```
danp: Ok how about this! Should what's here (trying all known passwords) happen every time Dial is called since the password the server wants could change while we're running? Based on my reading this will happen once for the lifetime of the pool/process.

apghero: That probably makes more sense, generally speaking. If the connection has been "AUTH"'d successfully, and the password is changed _online_, I don't think the connection will be broken, but the next Dial won't get the right password. So, yeah, this makes sense. I'll update.

apghero: Alright. The compromise here is to try first the last one we knew to be "good", falling back to the alternatives. When the "good" password is no longer valid, we nominate a new good one by trying them all.

danp:
I think this is looking good! Have you had a chance to try it out?

apghero:
I've not done a whole lot of testing against a real redis, no.

apghero:
@danp added a test that can be run with a REDIS_URL, which just gets a new pool connection and never releases it for every operation, then changes the password to something new. If the password is different at the end of the test, then, we know we've successfully connected multiple times.

hredis/redigo/redigo.go
```diff
@@ -42,15 +44,73 @@ func WaitForAvailability(url string, d time.Duration, f WaitFunc) (bool, error)
...
 }
 
+// make this a var, so we can change this for testing.
+var redisDialURL = redis.DialURL
+
 // NewRedisPoolFromURL returns a new *redigo/redis.Pool configured for the supplied url
 // The url can include a password in the standard form and if so is used to AUTH against
 // the redis server
-func NewRedisPoolFromURL(url string) (*redis.Pool, error) {
+func NewRedisPoolFromURL(rawURL string, altPasses ...string) (*redis.Pool, error) {
```
freeformz: What do you think about taking option funcs instead of altPasses and making these altPasses an optionFunc? If that were done then redisDialURL could also be specified with an optionFunc and we could do away with the redisDialURL for testing purposes.

hredis/redigo/redigo.go
```diff
@@ -42,15 +44,73 @@ func WaitForAvailability(url string, d time.Duration, f WaitFunc) (bool, error)
...
+		password string
+	)
+
+	// strip and save password
+	if pass, ok := u.User.Password(); ok {
+		password = pass
+		altPasses = append(altPasses, pass)
+	}
+
+	// DialURL will error if wrong password is set. We want to create a
```
freeformz: If possible I'd like it if code comments were 'de-personalized'.

hredis/redigo/redigo.go
```diff
@@ -42,15 +44,73 @@ func WaitForAvailability(url string, d time.Duration, f WaitFunc) (bool, error)
...
+		return nil, err
+	}
+
+	var (
+		mu       sync.Mutex
+		password string
+	)
+
+	// strip and save password
+	if pass, ok := u.User.Password(); ok {
```
freeformz: What do you think about using a single slice of passwords instead of a password and an optional slice of extra passwords. I believe using just a single slice would make some of the logic a little cleaner. I could be wrong though.

hredis/redigo/redigo.go
freeformz: The more I think about this, the more I'd much prefer to use option funcs for this.

hredis/redigo/redigo.go
apghero: I'm all for the simplicity of a single slice of passwords, but the common usage here is that the URL contains the password (e.g. on Heroku), which means that _someone_ will have to do this work...

hredis/redigo/redigo.go
apghero: I had thought about this, but it's a bit strange as we return a `*redis.Pool`, not something we can hide state in. The redisDialURL option and the passes option both change the behavior of the `redis.Pool.Dial` function, which means we'd need to create some intermediary to run the option functions on, in order to produce an appropriate method to use for the Dial function, or a closure. I think this approach yields a lot more complexity, but I'm willing to give it a shot.

apghero:
@freeformz I don't like how this came out, but here you go. Thoughts?

hredis/redigo/redigo.go
```diff
@@ -42,16 +45,129 @@ func WaitForAvailability(url string, d time.Duration, f WaitFunc) (bool, error)
 	}
 }
 
+// DialURLFunc describes the type of redis.DialURL, useful for changing the behavior
+// of a redis Dialer.
+type DialURLFunc func(string, ...redis.DialOption) (redis.Conn, error)
+
+// RedisDialer is an internal helper providing custom Redis dial logic for
```
freeformz: If it's internal, why is it exported?

hredis/redigo/redigo.go
freeformz: I was trying to suggest internally using a slice of passwords. In the common case there would only be one element in the slice. If there are 2 alternate passwords then there would be 3 elements in the slice. This way the dial() function could range on the slice and not have to track a seperate password.

hredis/redigo/redigo.go
apghero: I was thinking that in order to use `WithPasswords` you'd need the type exported, as part of the type is a function that exposes it. But, I'm likely wrong here.

apghero: Yup. I'm wrong. :)

apghero:
@freeformz your changes do clean up some stuff that bothered me. thanks! I'll incorporate some for sure.

apghero:
@freeformz I just included them all. Works for me.

PR #61, opened 2019-05-21
Run `go mod tidy`

freeformz:
Looks like `go mod tidy` wasn't run before committing changes previously


bernerdschaefer:
Would it make sense to run `go mod tidy` in CI and fail if it produces changes? For example, on our team's repos we run `dep check` as one of the stages of CI to ensure all dependencies are up-to-date before merging.

Otherwise 👍

freeformz:
@bernerdschaefer Probably.

PR #71, opened 2019-09-04
testenv: improvements on defaults

cyx:
## Context

Currently it was pointed to be a bit paranoid and try and avoid any
local installs.

While that's great for certain setups, that's not a good default for a
typical simple setup as well.

These new defaults allow a docker setup still, but also works well with
a typical brew install of redis / postgres.

### DIFF
--- a/testing/testenv/testenv.go
+++ b/testing/testenv/testenv.go
@@ -18,7 +18,7 @@ var skipMissingEnv = os.Getenv("TESTING_SKIP_MISSING_ENV") == "true"
 func NewRedisPool(t testing.TB) *redis.Pool {
 	t.Helper()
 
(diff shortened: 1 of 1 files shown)


cyx:
> If someone is going the docker path, you'd have to be running postgres/redis containers on their default ports? I assume that would be an issue if you also had a brew installed postgres/redis (which I imagine most people do) since they'd be competing for the same ports?

Yep, if they want to ensure the docker container can run alongside a host install, they'd want to bind the port to something other than the default yeah -- in which case I'd expect them to do:

```
DATABASE_URL=... REDIS_URL=... make test
```

In either case, this PR really is about that smart default for the host scenario.

PR #77, opened 2019-09-06
cmdutil: a bunch of common helpers around cmd

cyx:
## Context

We've went with a complete dependency injection model over the years,
making sure that cmd config is isolated well enough from the underlying
application / library code in terms of injecting dependencies.

These are some of the things we've added over the course of time.

Some documents are definitely needed.


go-kit/cmdutil/redispool/redispool.go
```diff
@@ -0,0 +1,62 @@
+package redispool
```
apghero: Doesn't this already exist as: https://github.com/heroku/x/blob/master/hredis/redigo/redigo.go#L103 and that package?

cyx: It kind of does, but then there's a bunch of gaps that are harder to reconcile.

freeformz: Style wise, I'm pretty -1 on this package.

Why?

library style package shouldn't be concerning themselves with dealing with configuration from environment variables. That is the domain of the main package to decide how to handle environment variables. This is something detailed in the design guide.

Everything is done against a Config type. That isn't idiomatic. I feel that `redispool.New(url string, ... other options)` would be way more idiomatic.

The attachment stuff isn't documented well enough for me to understand it enough to comment further.

freeformz: With that said, If others are +1 then I won't piss a fit.

go-kit/cmdutil/debug/config.go
```diff
@@ -0,0 +1,6 @@
+package debug
```
freeformz: please drop `go-kit`

go-kit/cmdutil/redispool/redispool.go
freeformz: If accepted, I plan to PR restructurings to these packages as I personally find them un-idiomatic.

cyx: If the restructuring is just how it's laid out, seems ok. If the restructuring is such that we have to drop the `env` bits, then the library would pretty much be useless from our perspective.

The primary reason for having the `Config` bits is that we reuse the same thing over and over across many services. Typing out the same `Config` struct or copy/pasting across many services would probably be not as great either.

I'm open to better ideas / suggestions which solves the underlying problem described above 👍

go-kit/cmdutil/debug/config.go
cyx: 👍 done

go-kit/cmdutil/redispool/redispool.go
freeformz: I will look into PR'ing ways around that.

PR #78, opened 2019-09-06
logplex: relaxing parser

lxfontes:
Replaces influxdata/go-syslog with a more relaxed parser.

Why not lpx?
- We are dealing with 2 formats: logplex ingress & egress. lpx only supports egress (drain)
- lack of prival parsing ( priority & version )
- no control over buffer allocations
  - meaning: one could stream 1Gb messages and there is no way to block that

How it works?
We decode in 2 layers:
- octet framing: built on top of [bufio](https://golang.org/pkg/bufio/#Scanner.Split) and finds message boundaries. Here we also enforce a max message size
- syslog: parse syslog fields & converts strings to objects ( time, integers, etc )

Having it as draft until we stress test it a bit more


freeformz:
Yes, please ditch lpx.

We should take a look at what *other* Go packages we have that use lpx and make sure this package work for them. Maybe even do the work to move them over to it?

lxfontes:
@freeformz that makes 💯 sense.  According to a quick [search](https://github.com/search?q=org%3Aheroku+%22github.com%2Fbmizerany%2Flpx%22&type=Code), we have ~ 50 spots using lpx. Thinking of:
- Send email notifying we are doing PRs wrt lpx
  - Give it a couple days cause #email
  - Codeowners that rather stay with lpx can reply to the email ( and hopefully tell us why )
- We send PRs to affected code
  - Make sure tests are passing ( update tests if needed )
  - Codeowners are responsible for merging / deploying at their discretion / schedule

how does that sound?

freeformz:
@lxfontes That sounds awesome.

PS: You may want to take a look at how [log-shuttle generates logplex](https://github.com/heroku/log-shuttle/blob/master/logplex_formatter.go) compatible messages. It's a bit of a hack, but it's generally philosophy is to generate as little garbage as possible. I'd be happy to go over it with you if you have questions.

PR #79, opened 2019-09-06
CircleCI / GolangCI-lint setup

freeformz:


tools.go
```diff
@@ -0,0 +1,7 @@
+// +build tools
+
+package tools
+
+import (
+	_ "github.com/golangci/golangci-lint/cmd/golangci-lint" // CLI mega lint tool
```
gaffneyc: This import line means that _any_ application using go modules which imports a subpackage from this repository will have the full golangci-lint command as a build dependency.

Why not just install it during CI?

freeformz: Yes, this is a problem. We're used to writing apps where the `tools.go` pattern works better.

We run linting locally too and want to make sure everyone is using the same linter version (as well as other tools).

freeformz: I take that back. I don't think that is the issue atm. The issue I believe is that versions of golangci-lint < 1.18.0 have broken / bad module info.

freeformz: I bumped the version of golangci-lint to v1.18.0 so the module problems should be resolved.

PR #85, opened 2019-09-06
tickgroup: add initial readme

agnaite:

We want to document the packages in `x/`.

## Description of changes

* Add a README with an overview and an example


apghero:
Would it make sense for this to actually be godoc, in say, package.go?

freeformz:
@agnaite Can we please convert this to godoc with examples? I'd be happy to help do so if you want. Why: godoc is the first place I go to find out how to use a package, including finding examples. This is pretty common in the go community.

PR #92, opened 2019-09-10
grpc: Re-adding Request-ID middleware

lxfontes:
This was removed in #77 and is needed by runtime.
@cyx: heads up in case this was removed on purpose.


cyx:
@lxfontes good catch!

PR #93, opened 2019-09-10
Add mustcert doc

jntullo:
Adds a godoc example for mustcert

### DIFF
--- a/testing/mustcert/example_test.go
+++ b/testing/mustcert/example_test.go
@@ -1,0 +1,73 @@
+package mustcert_test
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"io/ioutil"
+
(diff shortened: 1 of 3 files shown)


testing/mustcert/example_test.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	server.StartTLS()
+	defer server.Close()
+
+	// Create the Client configuration
+	cert, err := tls.X509KeyPair([]byte(clientCert.CertPEM()), []byte(clientCert.KeyPEM()))
+	if err != nil {
+		fmt.Println(err)
+	}
+	caCertPool := x509.NewCertPool()
```
lxfontes: are `Pool(ca.TLS())` and `caCertPool.AppendCertsFromPEM([]byte(ca.CertPEM()))` analogous?
if so, would keep the same style to not confuse the reader as to why is one different to another

jntullo: good call!

PR #102, opened 2019-10-28
logplex/encoding: Adding fallback time format

lxfontes:
Some of our clients are not sending a timezone neither `Z` as part of the their timestamps.
This used to be accepted by Logplex, so we are adding it back as fallback.


lxfontes:
thanks @freeformzSFDC 

we've been able to chase down the client and fixed it there 🙌  Given it was the only client error'ing, we feel this is not needed anymore.

freeformzSFDC:
SGTM.

PR #104, opened 2019-10-31
redispool: Add optional config to set max conn lifetime

ypaq:
## Rationale

We have seen leaking connection counts when missing to close redis  connections after `Get` from a pool. The redigo pool supports a `MaxConnLifetime` config 
(see https://github.com/gomodule/redigo/blob/master/redis/pool.go#L160-L162) which when set to non zero automatically closes connections older than the set duration.

## Changes
Here we introduce a new environment variable `REDIS_MAX_CONN_LIFETIME` which defaults to `0s` to retain the original behaviour.

### DIFF
--- a/cmdutil/redispool/redispool.go
+++ b/cmdutil/redispool/redispool.go
@@ -1,3 +1,5 @@
+// Package redispool supports setting up redis connection pools parameterized
+// via environment variables.
 package redispool
 
 import (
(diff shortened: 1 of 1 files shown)


cmdutil/redispool/redispool.go
```diff
@@ -48,9 +49,10 @@ func newPool(cfg Config) *redis.Pool {
 			}
 			return conn, nil
 		},
-		MaxIdle:     cfg.MaxIdleConns,
-		MaxActive:   cfg.MaxActiveConns,
-		IdleTimeout: cfg.IdleTimeout,
+		MaxIdle:         cfg.MaxIdleConns,
+		MaxActive:       cfg.MaxActiveConns,
+		MaxConnLifetime: cfg.MaxConnLifetime,
+		IdleTimeout:     cfg.IdleTimeout,
```
lxfontes: `0s` meaning: no limit. 👍 👎 ?

ypaq: > // Close connections older than this duration. If the value is zero, then
> // the pool does not close connections based on age.
> MaxConnLifetime time.Duration

freeformzSFDC: Please document these options via godoc.

PR #105, opened 2019-11-05
Add Open Census agent exporting to the Service.

freeformzSFDC:
Don't start the open census agent exporter unless the Service is created
with EnableOpenCensusTracing. This allows for a level of opt-in.

All Open Census env variables are prefixed with `OC_`. Only a few
options are currently supported, but others could be added later.

Other cleanups were done for my own clarity while reading through the
code.


cmdutil/service/standard.go
```diff
@@ -79,6 +79,14 @@ func New(appConfig interface{}, ofs ...OptionFunc) *Standard {
 	s.Add(debug.New(logger, sc.Debug.Port))
 	s.Add(signals.NewServer(logger, syscall.SIGINT, syscall.SIGTERM))
 
+	if o.enableOpenCensusTracing {
```
danp: Could this be more automatic based on presence of config, similar to how rollbar integration in the logger is skipped if config is missed? Maybe explicit enabling is fine to start but would be nice to keep a more on-by-default setup in mind.

freeformzSFDC: On by default is a goal, but IMO it's too soon for that. Technically enabling it by default is also a breaking API change (although that probably doesn't matter atm either).

cmdutil/service/standard.go
```diff
@@ -110,43 +118,49 @@ func (s *Standard) Run() {
...
 	if dyno == "" {
-		return ""
+		return dyno
 	}
 	parts := strings.SplitN(dyno, ".", 2)
-	pt := parts[0]
-	if pt == "web" {
-		pt = "server"
+	if parts[0] == "web" {
+		parts[0] = "server" // TODO[freeformz]: Document why this is server
```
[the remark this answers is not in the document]
freeformzSFDC: IMO this should be changed to return the dyno process type and that's it. If a given service wants to s/web/service/  that service can use `CustomMetricsSuffix`. But that is something that should go into a different PR. It was a mistake for me to include these things in this PR already. :-(

PR #106, opened 2019-11-05
Fix OC Agent / document cmdutil.Server and related / add examples.

freeformzSFDC:
ocagent.Exporter.Start() returns immediately, which violates oklog/run.Group semantics, so use NewContextServer instead to block until something else brings down the Group.

Also updated the documentation for the various Server types and created some rudimentary examples in hopes that future readers won't fall into the same traps that I did.

### DIFF
--- a/cmdutil/server.go
+++ b/cmdutil/server.go
@@ -6,53 +6,56 @@ import (
 	"github.com/oklog/run"
 )
 
-// A Server can be run synchronously and return an error.
+// Server runs synchronously and returns any errors. The Run method is expected
+// to block until Stop is called or return an error asap. Servers are typically
+// used with oklog/run.Group, where the first Server.Run to return will cancel
+// the Group (regardless of the error returned). Use NewContextServer to create
+// a Server that can block on a Context until Stop is called.
 //
-// Servers are typically used with oklog/run.Group.
 // TODO[freeformz]: Document why Stop takes an error and what to do with is.
 type Server interface {
 	Run() error
 	Stop(error)
 }
 
-// ServerFunc is a function which implements the Server interface.
+// ensure ServerFunc implements Server
+var _ Server = ServerFunc(func() error { return nil })
+
+// ServerFunc adapts a function to the Server interface. This is useful to adapt
+// a closure to being a Server.
 type ServerFunc func() error
 
-// Run calls fn and returns any errors.
-//
-// It implements the Server interface.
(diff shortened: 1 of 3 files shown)


cmdutil/server.go
```diff
@@ -6,53 +6,56 @@ import (
 	"github.com/oklog/run"
 )
 
-// A Server can be run synchronously and return an error.
+// Server runs synchronously and returns any errors. The Run method is expected
+// to block until Stop is called or return an error asap. Servers are typically
```
danp:
```suggestion
// to block until finished, returning any error, or until Stop is called.
```

cmdutil/server.go
```diff
@@ -6,53 +6,56 @@ import (
...
 func (sf ServerFuncs) Stop(err error) {
 	if sf.StopFunc != nil {
 		sf.StopFunc(err)
 	}
 }
 
-// NewContextServer returns a Server that runs the given
-// function with a context that is canceled when the Server
-// is stopped.
+// NewContextServer that when Run(), calls the given function with a context
```
danp:
```suggestion
// NewContextServer returns a Server that when Run(), calls the given function with a context
```

Is this much different from what was there?

cmdutil/server.go
```diff
@@ -67,9 +70,9 @@ func NewContextServer(fn func(context.Context) error) Server {
 
 }
 
-// MultiServer returns a new kit.Server which will run all of the provided
-// servers until one of them fails or the server is stopped.
-func MultiServer(srvs ...Server) Server {
+// MultiServer which, when Run, will run all of the servers until one of them
```
danp:
```suggestion
// MultiServer returns a Server which, when Run, will run all of the servers until one of them
```

cmdutil/server.go
freeformzSFDC: I really do not like stating "returns an X" as that information is already present in the name "NewContextServer" (if it didn't return a Server of some type it is misnamed) and in the return type (it returns a Server).

danp: Hmm, okay, but it does read aloud very awkwardly.

freeformzSFDC: From my perspective "NewContextServer returns a Server" reads strangely out loud.

I read these as: "New Context Server returns a Server...", so yeah, why did you say "returns a Server", I already knew that.

PR #108, opened 2019-11-07
Wire up GRPC tracing via ocgrpc

freeformzSFDC:


grpc/grpcserver/options.go
```diff
@@ -67,6 +68,14 @@ func AuthInterceptors(unary grpc.UnaryServerInterceptor, stream grpc.StreamServe
 	}
 }
 
+// WithOCGRPCServerHandler sets the grpc server up with provided ServerHandler
+// as its StatsHandler
+func WithOCGRPCServerHandler(h *ocgrpc.ServerHandler) ServerOption {
```
danp: Is there anything OC-specific about this? Should it be WithStatsHandler?

Going further, some of the ServerOption funcs in this package exist to help wrap up setup that would need to be done in multiple places or require more involved repeated setup, such as chaining interceptors, handling both unary and streaming, etc. Is that true for setting the stats handler? If not, could this be done with:

```go
srv := grpcserver.New(grpcserver.GRPCOption(grpc.StatsHandler(h)))
```

PR #109, opened 2019-12-05
basicauth: Use constant time comparison in checker

ypaq:
## Rationale

String comparison can be susceptible to side-channel attacks.

## Changes

Use constant time string comparison.

### DIFF
--- a/hmiddleware/basicauth/checker.go
+++ b/hmiddleware/basicauth/checker.go
@@ -1,6 +1,7 @@
 package basicauth
 
 import (
+	"crypto/subtle"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


hmiddleware/basicauth/checker.go
```diff
@@ -63,9 +64,8 @@ func NewChecker(credentials []Credential) *Checker {
 // Valid is true if username and password represent acceptable credentials.
 func (c *Checker) Valid(username, password string) bool {
 	for _, cred := range c.credentials {
-		if cred.Username == username && cred.Password == password {
-			return true
-		}
+		return subtle.ConstantTimeCompare([]byte(cred.Username), []byte(username)) == 1 &&
```
danp: Maybe going too far, but there's still a short-circuit possibility here, I think. For example, if the username check returns something other than 1, then the password check never happens, which might influence the timing. Could rearrange this to always do both checks and then check the results.

lxfontes:
```
--- FAIL: TestChecker (0.00s)
    checker_test.go:99: 4. got false, want true
    checker_test.go:99: 5. got false, want true
```
the `if` is so we walk through all valid creds

ypaq: @lxfontes fixed the test. It was passing for empty username or empty password. I changed this behavior: you can't have an empty user or pass word which I think is ok for where we use basicauth.

ypaq: @danp addressed your comment with the latest commit.

ypaq: @lxfontes disregard my last comment. ;)

PR #112, opened 2019-12-13
cmdutil: Lossy logger

ypaq:
## Rationale

We want logging facilities that allow us to limit the amount of logs a system can produce. Here we implement a SampleLogger that ~implements `Printf` given a logger and a limit~ rate-limits logs per window. It drops logs that exceed this limit per window.
This logger also skips hooks to avoid spamming 3rd party telemetry.

## Changes

Add SampleLogger.


ypaq:
@lxfontes do you think there's value in implementing the rest and support logrus's functions entirely?

lxfontes:
I think using `Printf` is an OK compromise that doesn't fiddle with log levels. 🚀

freeformzSFDC:
FWIW: If this is meant to be used with logrus, I would probably choose to do this like so: https://play.golang.org/p/bqb5k85DX9C

This would allow the user to utilize any logrus logging method.

freeformzSFDC:
My example implementation also side steps having to use Printf, which isn't how most people use logrus.

lxfontes:
@freeformz while that example does "print" the correct output, the entire logrus hook pipeline is still exercised. meaning it would spam components like rollbar, which is exactly what we are trying to avoid

danp:
@lxfontes I _think_ what @freeformz was getting at is related to your comment:

> callers have to explicitly call Printf

Instead of basing SampleLogger around Printf, which people using logrus typically don't use (in my experience), basing on logrus.Formatter as in @freeformz's example makes it work for all logrus methods, such as Info, Error, etc.

lxfontes:
@danp i understood his point.
what i'm saying is neither solutions are good here, and in fact we need to impersonate `logrus.FieldLogger`.

lxfontes:
@danp @ypaq pls check if this approach makes sense.

danp:
I think so! To confirm: is it the intention that using this logger would never use hooks set on the original? Like, should rollbar *never* be activated via a hook on the sampled logger?

lxfontes:
yes! we want to capture high throughput errors ( ex: logplex ingestion & drain errors ), that are guaranteed to be spammy but useful for debugging.
something like:

```
err := doDrain()
if err != nil {
 metrics.inc(err)
 lossyLogger(err)
}
```

that metric has a high value, but somewhat stable.
so once we see the metrics going 📈 , we can take a look at the sampled error log and have a hint of what is going on.
it avoids a re-deploy with debug=on & keeps some trailing history for context.

PR #113, opened 2020-01-09
Import loggingtags

bernerdschaefer:
The Runtime team has an internal protobuf plugin called `protoc-gen-loggintags` which allows protobuf message fields to be annotated to mark whether they're safe to log. The plugin codegens a `LoggingTags() map[string]interface{}` method on the messages.

When using the `grpcserver` package here in heroku/x, messages which implement this method will have request and response fields marked as safe automatically included in the gRPC context tags and be included in logs and traces.

This imports runtime's plugin for use by other teams and services and adds documentation for how to use it.


cmd/protoc-gen-loggingtags/Makefile
```diff
@@ -0,0 +1,43 @@
+TOP_LEVEL = $(shell git rev-parse --show-toplevel)
```
bernerdschaefer: Should the Makefile be rolled into the toplevel Makefile or kept separate as here?

cmd/protoc-gen-loggingtags/README.md
```diff
@@ -0,0 +1,54 @@
+# protoc-gen-loggingtags
```
bernerdschaefer: Do we want this in heroku/x or its own repo?

cmd/protoc-gen-loggingtags/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+The `grpcserver` package automatically checks if request or response messages
+implement the LoggingTags method and includes the returned tags in the context
+tags for logging.
+
+## Example
+
+```go
+syntax = "proto3";
+import "github.com/heroku/x/cmd/protoc-gen-loggingtags/heroku/loggingtags/safe.proto";
```
bernerdschaefer: Runtime imports the proto as `heroku/loggingtags/safe.proto`, and then rewrites the import via flags on protoc. It's more common ([example](https://github.com/mwitkow/go-proto-validators#installing-and-using)) to specify the full package import and set the protoc path to GOPATH or vendor, but that produces a very long (and repetitive) import path here.

Perhaps we should reorganize now, so the import is something like `github.com/heroku/x/cmd/protoc-gen-loggingtags/safe.proto`?

cmd/protoc-gen-loggingtags/Makefile
danp: Couple things I'm thinking about: this would be the second command in x, behind s3env, and it seems likely most of what happens here would need to happen in each project that wants to use protoc-gen-loggingtags. Perhaps we should think about how we expect people to build and use this command?

cmd/protoc-gen-loggingtags/README.md
danp: I think in here is fine, unless it changes some of the answers to the Makefile questions.

cmd/protoc-gen-loggingtags/README.md
danp: Would it help anything to have the proto somewhere else in the tree other than under cmd/protoc-gen-loggingtags? Either way, probably ending up with whatever requires the least config when consuming would be preferable.

lxfontes:
agree with dan's comments and 💜 for doing this

cmd/protoc-gen-loggingtags/Makefile
bernerdschaefer: Assuming a project is already using gRPC and protobufs, what's required is that:

* `protoc-gen-loggingtags` is installed into your path (or the path to it is provided to `protoc`)
* The `safe.proto` file is available somewhere in source form, such that one of the paths passed to protoc as a `--proto_path` / `-I` flag + the path specified in the `import` statement in the proto file resolve to the `safe.proto` file. For example, `--proto_path=$GOPATH/src` and `import "github.com/heroku/x/a/b/c/safe.proto` or `--proto_path=$GOPATH/src/github.com/heroku/x` and import `a/b/c/safe.proto`.

I'd guess that users would vendor the command and install locally, which is presumably the same thing they're already doing for other tools.

danp: That all sounds good! I guess we need to get any testing running in CI and break out any bits around using in a project into docs?

bernerdschaefer: Okay, reflecting on all of this, I think we could go with something like:

* Detailed docs in cmd/x/protoc-gen-loggingtags
* Move the proto into a toplevel dir, so imports are simplified to `github.com/heroku/x/loggingtags/safe.proto`
* Move the build process into the toplevel Makefile
* Add brief info to toplevel README that links to the cmd's detailed docs

How does that sound?

danp: Sounds good!

bernerdschaefer:
I've reorganized based on the results of the discussion from https://github.com/heroku/x/pull/113#discussion_r366513865.

The individual commits since the previous iteration are relatively small, but I wanted to call out that there are now tests for the generator (mostly in 34d26d8) with lots of inspiration from how protoc-gen-go is tested. The tests cover which files are generated and also fails if it produces a diff for the test data sample.

The rest are mostly path fixes, doc changes, and merging the Makefile into the root.

bernerdschaefer:
Hmm.. there are [two CI linter failures](https://circleci.com/gh/heroku/x/217):

```
cmd/protoc-gen-loggingtags/internal/gen/generate.go:79:2: Consider preallocating `files` (prealloc)
cmd/protoc-gen-loggingtags/golden_test.go:34:16: struct of size 72 bytes could be of size 64 bytes:
```

I'm almost certain I don't care about either and that they should be informational rather than blocking, but I guess x is all or nothing on linters. I'll consider changing or throwing ignore annotations on them.

PR #114, opened 2020-01-14
Add logger middleware

jessieay:
* Pulled from https://github.com/heroku/runtime/blob/fad8ed0baa88b7bf86a4888cbabc4256c50886be/lib/middleware/logger_test.go
* We are going to use this in Lyra Identity soon and I saw that pulling
this into X was previously suggested when the same code was added to the
Identity prototype https://github.com/heroku/lyra-labs/pull/208/files#r343058787

### DIFF
--- a/hmiddleware/logger.go
+++ b/hmiddleware/logger.go
@@ -8,7 +8,7 @@ import (
 	"github.com/go-chi/chi"
 	"github.com/go-chi/chi/middleware"
 
-	httprequestid "github.com/heroku/x/requestid"
+	"github.com/heroku/x/requestid"
 
 	"github.com/sirupsen/logrus"
 )
@@ -51,7 +51,7 @@ func PostRequestLogger(l logrus.FieldLogger) func(next http.Handler) http.Handle
 
 func logRequest(l logrus.FieldLogger, r *http.Request, status int, bytes int, service time.Duration, at string) {
 	log := l.WithFields(logrus.Fields{
-		"request_id":  httprequestid.Get(r),
+		"request_id":  requestid.Get(r),
 		"method":      r.Method,
 		"host":        r.Host,
 		"path":        r.URL.RequestURI(),
--- a/hmiddleware/logger.go
+++ b/hmiddleware/logger.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/go-chi/chi"
 	"github.com/go-chi/chi/middleware"
 
 	"github.com/heroku/x/requestid"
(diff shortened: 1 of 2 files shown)


danp:
Maybe @bernerdschaefer could give this a quick look as well?

hmiddleware/logger.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+
+	httprequestid "github.com/heroku/x/requestid"
```
jessieay: I did this because it's what we are doing at https://github.com/heroku/x/blob/839ea2283620b86e226c885c0ac0cb30a3e6014e/grpc/grpchttp/request_id.go#L10

not sure if that is a good reason or not...

bernerdschaefer: The reason why it's renamed to `httprequestid ` there is because it's also importing a gRPC requestid package and the names would conflict: https://github.com/heroku/x/blob/839ea2283620b86e226c885c0ac0cb30a3e6014e/grpc/grpchttp/request_id.go#L9-L10

Using the default import name of `requestid` should be fine here. I've also made a note to see if it makes sense to unify (or reorganize) the two requestid packages to prevent the naming conflicts.

hmiddleware/logger.go
```diff
@@ -0,0 +1,87 @@
...
+		"path":        r.URL.RequestURI(),
+		"remote_addr": r.RemoteAddr,
+		"user_agent":  r.UserAgent(),
+		"protocol":    r.URL.Scheme,
+		"at":          at,
+	})
+
+	if r.Context().Value(chi.RouteCtxKey) != nil {
+		if id := chi.URLParam(r, "app_id"); id != "" {
+			log = log.WithField("appID", id)
```
bernerdschaefer: This bit which pulls out and logs app_id from the URL params if present was done for a specific service. We maybe don't need or want it in this more general context?

hmiddleware/logger.go
jessieay: 🆒 

reverted to use default package name at f3418ff

hmiddleware/logger.go
jessieay: good call, removed the code specific to `appID` at 5d4ba85

hmiddleware/logger.go
```diff
@@ -0,0 +1,80 @@
...
+		return http.HandlerFunc(fn)
+	}
+}
+
+func logRequest(l logrus.FieldLogger, r *http.Request, status int, bytes int, service time.Duration, at string) {
+	log := l.WithFields(logrus.Fields{
+		"request_id":  requestid.Get(r),
+		"method":      r.Method,
+		"host":        r.Host,
+		"path":        r.URL.RequestURI(),
```
mikehale: For midgard we try to log the templatized path, so `/apps/{param1}` instead of `/apps/foo`, but in our case we can do that because we are using mux for routing which knows about the path templates. I wonder if there is a generic way to handle that.

danp: There's code [in httpmetrics](https://github.com/heroku/x/blob/4b3828443f9722a6c2413023fe79a69c797f33d8/hmiddleware/httpmetrics/httpmetrics.go#L66-L100) that assumes use of chi and extracts route patterns for metrics naming. That could be used there and here to get the pattern/template, if desired.

PR #118, opened 2020-01-23
Re-work logging middleware to be compatible with Chi

jessieay:
* Previously, I added logging middleware that coudld be used by the Chi
router but did not interact with Chi.
* When we were testing our rollbar setup in Lyra Identity, we found that
the Panic recovery was not working within Chi router endpoints.
* This was because the Recovery middleware depends on the Logging
middleware that Chi provides, which we were not using because we were
using this handy Logging middleware from X instead.
* At this point, it seemed like a better solution than writing our own
recovery middleware for Chi was to implement the LogFormatter and
LogEntry interfaces so we can use the Chi middleware for logging and
panic recovery.
* This is the approach shown here https://github.com/go-chi/chi/blob/cca4135d8dddff765463feaf1118047a9e506b4a/_examples/logging/main.go
* Also used by Lyra Labs here https://github.com/heroku/lyra-labs/pull/212/
* Logging has been adapted to use the fields that other Heroku services
use for easier Splunking etc.

Thanks for the pairing help on this, @bernerdschaefer !


hmiddleware/request_logger.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Panic is called by Chi's Recoverer middleware.
+// See https://github.com/go-chi/chi/blob/baf4ef5b139e284b297573d89daf587457153aa3/middleware/recoverer.go
+func (l *StructuredLoggerEntry) Panic(v interface{}, stack []byte) {
+	p := fmt.Sprintf("%+v", v)
+	l.Logger.WithFields(logrus.Fields{
+		"stack": string(stack),
+		"panic": p,
+	}).Panic()
```
bernerdschaefer: Note that calling `Panic` on the logger here will actually cause the handler to re-panic, rather than gracefully handling the panic with a 500. I think we'll want this to be something like `.WithError(v).Error("unhandled panic")` as in the [grpc panic handler](https://github.com/heroku/x/blob/master/grpc/panichandler/panichandler.go#L19).

It looks like you could also do `WithError(v).Log(logrus.PanicLevel, "unhandled panic")` if you wanted the panics to be uniquely identified by that log level (but without re-panicing like `Panic(msg)` does.

hmiddleware/request_logger_test.go
```diff
@@ -0,0 +1,50 @@
...
+	status := 200
+	bytes := 0
+	elapsed := time.Duration(100)
+
+	e.Write(status, bytes, elapsed)
+}
+
+func TestLogEntryPanic(t *testing.T) {
+	req, _ := http.NewRequest(http.MethodGet, "/", nil)
+	logger := logrus.New()
```
bernerdschaefer: This is a good case for using the [testlog package](https://godoc.org/github.com/heroku/x/testing/testlog#NewNullLogger). It looks like there aren't any examples there, but your test might look something like:

```go
logger, hook := testlog.NewNullLogger()
l := &StructuredLogger{
	Logger: logger,
}
e := l.NewLogEntry(req)

err := errors.New("test error")
stack := "foo.go:123"
e.Panic(err, []byte(stack))

hook.CheckContained(t, `error="test error"`, `stack="foo.go:123"`)
```

jessieay: ah yeah, I was using that for the old logger middleware that was removed in this PR.

@jboursiquot-sfdc's opinion was that testing what the logs themselves contain was looking too much at the implementation details of the methods we are adding (rather than the output of the methods) but it does feel odd to not have any tests looking at the log output. 🤔

danp:
Can we perhaps leave the existing logging middlewares for backwards compat and use not with chi? I think at least some places use them currently but with standard net/http stuff and not chi.

jessieay:
@danp I am open to leaving the existing middleware! I wasn't sure if anyone other than us was using it. But since X is open source, it's possible there are external parties using it.

My only hesitation with keeping both around is that it means we are providing 2 different paths for logging with Chi and it would be nice to clearly lay out a "golden path" for logging / panic handling.

hmiddleware/request_logger.go
jessieay: makes sense, I thought perhapsI was doing a special logger Panic here and it would just log but I could tell from the test that the panic was not being recovered.

Updated at c78396d to make it more similar to the grpc panic handler you linked, I think hat makes a lot of sense.

jessieay:
@danp added back in at 3c69a39

hmiddleware/request_logger_test.go
jboursiquot-sfdc: @jessieay Testing the log output here does make sense since we're checking that `e.Panic` should emit certain key/value pairs. The one where it wasn't warranted was for the `NewLogEntry` test since we were only checking that a `LogEntry` was indeed returned.

jessieay: note to self: look into `hook.LastEntry` so you can look and see if the fields match what you expect

jessieay: added log output info in specs at a67c99e

jessieay:
@jboursiquot-sfdc added some godoc comments here at 27e28ba to hopefully make the existence of 2 logging middlewares less confusing

PR #119, opened 2020-01-25
cmdutil/service: support configuring HTTP server

bernerdschaefer:
Go's HTTP server supports many additional configurations beyond those
managed automatically by the standard HTTP service. Some users want to
be able to set these additional settings without losing the defaults.

This adds a `WithHTTPServerHook` function to allow users to hook into
the server setup process and set additional configuration options before
the server is started.


cmdutil/service/standard_http_test.go
```diff
@@ -6,8 +6,12 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"os"
 	"testing"
 
+	"github.com/jkakar/envtest"
```
lxfontes: IMO i don't see much value in importing a package where a couple of functions manipulating the environment would do.

bernerdschaefer: Fair enough, I knew we had used envtest in some other projects to deal with tests leaking environment variables, but I see that this repo has a few other cases where they're directly cleaned up w/ the os package. I've updated with f4697806a.

go.mod
```diff
@@ -4,7 +4,6 @@ go 1.12
 
 require (
 	contrib.go.opencensus.io/exporter/ocagent v0.6.0
-	github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d // indirect
```
bernerdschaefer: `go mod tidy` 🤷‍♂

PR #122, opened 2020-01-30
Add appended request_id to headers

jessieay:
* If there is a X-Request-Id header, it was being added to the context but
not the headers
* Headers are what gets logged by the request_logger middleware so
adding to headers is helpful for logging/splunking
* https://gus.my.salesforce.com/a07B00000078HcyIAE

### DIFF
--- a/hcontext/requestid.go
+++ b/hcontext/requestid.go
@@ -8,6 +8,7 @@ package hcontext
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


hcontext/requestid.go
```diff
@@ -26,14 +26,16 @@ var headersToSearch = []string{
 // If one is found, it appends a new request ID and sets the comma separated value as the header.
 // If one is not found, it sets a new request ID as the header.
 func FromRequest(r *http.Request) (id string, ok bool) {
+	newRID := uuid.New().String()
+
 	for _, try := range headersToSearch {
 		if id = r.Header.Get(try); id != "" {
-			newRequestID := uuid.New().String() + "," + id
-			return newRequestID, true
+			newRID = newRID + "," + id
```
jessieay: @jboursiquot-sfdc is this valid go?

jboursiquot-sfdc: Yes, though it would be more idiomatic to do something like this: `newRID = fmt.Sprintf("%s,%s", newRID, try)`

jessieay:
@jboursiquot-sfdc ready for re-review!

PR #123, opened 2020-01-31
Tags middleware to annotate HTTP request log lines

dmcinnes:
A @bernerdschaefer and @dmcinnes joint

This adds a new middleware called `Tags` which allows annotation of the HTTP log lines with StructuredLogger by adding tag data into the Context.

Additionally, this fixes a bug with the `CheckContained` function in testlog which wasn't properly checking lists of arguments.

This uses the `github.com/grpc-ecosystem/go-grpc-middleware/tags` library for the tag handling. Even though it comes from this library, the functionality is independent of gRPC.


go.mod
```diff
@@ -11,13 +11,13 @@ require (
...
 	github.com/go-ini/ini v1.33.0 // indirect
-	github.com/go-kit/kit v0.6.0
+	github.com/go-kit/kit v0.9.0
 	github.com/gogo/protobuf v1.3.0 // indirect
 	github.com/golang/protobuf v1.3.2
 	github.com/gomodule/redigo v2.0.0+incompatible
 	github.com/google/gops v0.3.7
 	github.com/google/uuid v1.1.1
-	github.com/grpc-ecosystem/go-grpc-middleware v1.0.0
+	github.com/grpc-ecosystem/go-grpc-middleware v1.1.1-0.20200128160657-3c51f7f33212
```
bernerdschaefer: The last release of go-grpc-middleware was a few months ago, and doesn't include the tags functions we needed, so this is pointed at master.

bernerdschaefer: https://github.com/grpc-ecosystem/go-grpc-middleware/issues/259

PR #125, opened 2020-02-06
testing/testlog: add New function

bernerdschaefer:
The testlog package's function to get a test logger was `NewNullLogger`,
which was used to match the function name in logrus' hook test package.
Most folks using this package now, however, have not used the logrus
package, and there are no other kinds of test loggers in the package, so
using `New` instead reads better.

This also documents the package and adds examples.

### DIFF
--- a/testing/testlog/example_test.go
+++ b/testing/testlog/example_test.go
@@ -1,0 +1,28 @@
+package testlog_test
+
+import (
+	"testing"
+
+	"github.com/heroku/x/testing/testlog"
+)
+
+func Example() {
+	var t *testing.T // use t provided by test function
+
+	logger, hook := testlog.New()
+
+	// Use the test logger
+	logger.WithField("a", "b").Info("test")
+
+	hook.CheckAllContained(t, "a=b", "msg=test")
+	hook.CheckNotContained(t, "unexpected")
+}
+
+func Example_discard() {
+	logger, _ := testlog.New()
+
+	// Use the test logger
+	logger.WithField("a", "b").Info("test")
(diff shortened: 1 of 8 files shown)


testing/testlog/testlog.go
```diff
@@ -17,16 +18,23 @@ type Hook struct {
 	entries []*logrus.Entry
 }
 
-// NewNullLogger Creates a discarding logger and installs the test hook.
-func NewNullLogger() (*logrus.Logger, *Hook) {
+// New sets up a test logger that produces no output. Use the returned hook to
+// observe and make assertions about what was logged.
+func New() (*logrus.Logger, *Hook) {
```
bernerdschaefer: I also considered updating this signature to only return a single thing which implements both hook and logger functionality (likely embedding `*logrus.Logger` into `Hook`), so that it's not necessary to explicitly choose whether you want to discard logs or test their output.

That said, I do think it's a bit clearer this way, if less convenient.

testing/testlog/example_test.go
```diff
@@ -0,0 +1,28 @@
+package testlog_test
+
+import (
+	"testing"
+
+	"github.com/heroku/x/testing/testlog"
+)
+
+func Example() {
+	var t *testing.T // use t provided by test function
```
bernerdschaefer: This is a bit awkward, since there's no way to get a proper T here in the same way an end-user would. One option would be to omit these testable examples and put them into plain docs (which is how the [testing package](https://golang.org/pkg/testing/) shows examples).

PR #130, opened 2020-03-13
Adding Timeout Configs to config.go

brompwnie:
Adding Read and Write timeout properties. These values can serve as failsafes should the component's load balancer i.e ELB no longer implement timeouts for some reason. This is a defence-in-depth measure.

### DIFF
--- a/cmdutil/https/config.go
+++ b/cmdutil/https/config.go
@@ -1,5 +1,7 @@
 package https
 
+import "time"
+
 // Config for HTTP and HTTPS servers.
 type Config struct {
 	InsecurePort int `env:"HEROKU_ROUTER_HTTP_PORT,required"`
@@ -7,7 +9,9 @@ type Config struct {
 
 	// These environement variables are automatically set by ACM in
 	// relation to Let's Encrypt certificates.
-	ServerCert  string `env:"SERVER_CERT,required"`
-	ServerKey   string `env:"SERVER_KEY,required"`
-	UseAutocert bool   `env:"HTTPS_USE_AUTOCERT"`
+	ServerCert   string        `env:"SERVER_CERT,required"`
+	ServerKey    string        `env:"SERVER_KEY,required"`
+	UseAutocert  bool          `env:"HTTPS_USE_AUTOCERT"`
+	ReadTimeout  time.Duration `env:"READ_TIMEOUT"`
+	WriteTimeout time.Duration `env:"WRITE_TIMEOUT"`
 }
--- a/cmdutil/https/config.go
+++ b/cmdutil/https/config.go
@@ -7,11 +7,12 @@ type Config struct {
 	InsecurePort int `env:"HEROKU_ROUTER_HTTP_PORT,required"`
 	SecurePort   int `env:"HEROKU_ROUTER_HTTPS_PORT,required"`
 
-	// These environement variables are automatically set by ACM in
+	// These environment variables are automatically set by ACM in
 	// relation to Let's Encrypt certificates.
-	ServerCert   string        `env:"SERVER_CERT,required"`
-	ServerKey    string        `env:"SERVER_KEY,required"`
-	UseAutocert  bool          `env:"HTTPS_USE_AUTOCERT"`
-	ReadTimeout  time.Duration `env:"READ_TIMEOUT"`
-	WriteTimeout time.Duration `env:"WRITE_TIMEOUT"`
+	ServerCert  string `env:"SERVER_CERT,required"`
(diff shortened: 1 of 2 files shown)


cmdutil/https/serve.go
```diff
@@ -20,8 +21,10 @@ func Serve(handler http.Handler, cfg Config, opts ...func(*http.Server)) error {
 		serverCert = []byte(cfg.ServerCert)
 		serverKey  = []byte(cfg.ServerKey)
 		srv        = &http.Server{
-			Addr:    fmt.Sprintf(":%d", cfg.SecurePort),
-			Handler: handler,
+			Addr:         fmt.Sprintf(":%d", cfg.SecurePort),
+			Handler:      handler,
+			WriteTimeout: cfg.WriteTimeout * time.Second,
```
danp: If WriteTimeout and ReadTimeout are already time.Duration the multiply by time.Second can be skipped.

cmdutil/https/config.go
```diff
@@ -1,13 +1,17 @@
...
 
 	// These environement variables are automatically set by ACM in
 	// relation to Let's Encrypt certificates.
-	ServerCert  string `env:"SERVER_CERT,required"`
-	ServerKey   string `env:"SERVER_KEY,required"`
-	UseAutocert bool   `env:"HTTPS_USE_AUTOCERT"`
+	ServerCert   string        `env:"SERVER_CERT,required"`
+	ServerKey    string        `env:"SERVER_KEY,required"`
+	UseAutocert  bool          `env:"HTTPS_USE_AUTOCERT"`
+	ReadTimeout  time.Duration `env:"READ_TIMEOUT"`
```
danp: Could we perhaps group these separately from the ACM/LE-related things here? And should they have more specific names, such as HTTP_SERVER_READ_TIMEOUT?

brompwnie: @danp do you have a suggestion in terms of the grouping? I stuck with ReadTimeout and Writetimeout to keep parity with the actual property names. These are TCP timeouts essentially that are applied in the context of HTTP so by naming them HTTP_SERVER_READ_TIMEOUT, would that still be accurate? Happy to take suggestions here and follow any best practices, still getting to grips with Go :)

cmdutil/https/serve.go
brompwnie: Ah yes good catch, I'll address that.

cmdutil/https/config.go
bernerdschaefer: Note the comment above, which says, "These environement variables are automatically set by ACM in relation to Let's Encrypt certificates." That makes it seem like these configurations are also part of the ACM config. It would be clearer if you threw in a newline and a separate comment for the timeout configurations.

+1 to calling the environment variable HTTP_SERVER_READ/WRITE_TIMEOUT. ReadTimeout and WriteTimeout are fine for the struct field names since it's clear from context.

cmdutil/https/config.go
```diff
@@ -1,13 +1,17 @@
...
+	WriteTimeout time.Duration `env:"WRITE_TIMEOUT"`
```
bernerdschaefer: Are there defaults we could apply here that represent best practices or upper bounds? For example, something that matches how we configure ELBs or the timeouts we use in the dogwood router?

Definitely not a blocker, making it straightforward to configure is a great start (and less risky), but it would be great to eventually land somewhere that doesn't require everyone to remember to configure this themselves.

brompwnie: @bernerdschaefer I've been investigating our ELB (Cedar) values and they use the AWS default of 60s, defaulting to them would be a good start or the Dogwood values? I'm not aware of any best practices around this, do you have any suggestions?

cmdutil/https/config.go
brompwnie: Oh yes that totally makes sense, I'll definitely add that comment to separate the values.

+1 for calling the env vars HTTP_SERVER_READ/WRITE_TIMEOUT.

cmdutil/https/config.go
bernerdschaefer: Screen Shot 2020-03-17 at 10 30 20 AM

Looks like Dogwood also uses the default. I'm 👍 on defaulting the read/write timeouts to 60s to start with. If you don't know, you can do that by updating the struct tag to on the fields to be `env:"HTTP_SERVER_WRITE_TIMEOUT,default=60s"`.

brompwnie: @bernerdschaefer ack and thanks 👍

brompwnie:
Thanks for the assistance @bernerdschaefer @danp 🎆

PR #132, opened 2020-03-24
encoding/logplex: validating prival field

lxfontes:
Some clients send malformed syslog messages that can panic logfwds.
this protects against those malformed messages by checking if the regex had a full match  before using prival.


logplex/encoding/scanner_test.go
```diff
@@ -103,6 +103,14 @@ func TestScanner(t *testing.T) {
...
 			err: bufio.ErrTooLong,
 		},
+
+		"bad prival": {
+			log:          "64 <190>11 2019-07-20T17:50:10.879238Z shuttle token shuttle - - 99\n 65 <190>1 2019-07-20T17:50:10.879238Z shuttle token shuttle - - 10",
+			count:        1,
+			err:          ErrBadFrame,
```
iamjem: wouldn't we see `ErrInvalidPriVal`?

lxfontes: you're right. fixed the test

PR #133, opened 2020-03-30
hmiddleware: add connection closing middleware

iamjem:
We'd like the ability to cooperatively close connections once we've received a certain number of requests, to avoid the pinning of high traffic connections to a particular server process.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/google/uuid v1.1.1
 	github.com/grpc-ecosystem/go-grpc-middleware v1.2.0
 	github.com/grpc-ecosystem/grpc-gateway v1.9.6
+	github.com/hashicorp/golang-lru v0.5.1
 	github.com/heroku/rollrus v0.1.1
 	github.com/hydrogen18/memlistener v0.0.0-20141126152155-54553eb933fb
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
--- a/hmiddleware/connection_closer.go
+++ b/hmiddleware/connection_closer.go
@@ -27,7 +27,7 @@ func ConnectionClosingContext(ctx context.Context, c net.Conn) context.Context {
 // ConnContext field.
 func ConnectionClosingMiddleware(mp metrics.Provider, maxRequests, cacheSize int) (func(http.Handler) http.Handler, error) {
 	var (
-		closed = mp.NewCounter("server.connection.closes.total")
+		closed = mp.NewCounter("connection.closes.total")
 		mtx    sync.Mutex
 	)
 
--- a/hmiddleware/connection_closer.go
+++ b/hmiddleware/connection_closer.go
@@ -1,0 +1,75 @@
+package hmiddleware
+
+import (
+	"context"
+	"log"
+	"net"
+	"net/http"
+	"sync"
+
+	"github.com/google/uuid"
+	"github.com/hashicorp/golang-lru"
+	"github.com/heroku/x/go-kit/metrics"
+)
+
+type connCloseKey int
+
+var connCloseID connCloseKey
+
+// ConnectionClosingContext adds a unique identifier to the context for a newly established connection. This function
+// is meant to be used as an http.Server's ConnContext field, and is required for the ConnectionClosingMiddleware.
+func ConnectionClosingContext(ctx context.Context, c net.Conn) context.Context {
+	return context.WithValue(ctx, connCloseID, uuid.New().String())
+}
+
+// ConnectionClosingMiddleware returns middleware that will add a "Connection: close" responses header after a single
+// connection has produced maxRequests http requests. The cacheSize determines the size of the LRU cache which tracks
+// connection request counts. This middleware requires the http.Server sets ConnectionClosingContext for it's
+// ConnContext field.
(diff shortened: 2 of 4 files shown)


hmiddleware/connection_closer.go
```diff
@@ -0,0 +1,75 @@
...
+				}
+
+				mtx.Unlock()
+			}
+
+			next.ServeHTTP(w, r)
+
+			if shouldClose {
+				closed.Add(1)
+				w.Header().Add("connection", "close")
```
danp: If I understand correctly, this can only be called here if next didn't write anything to w. Should this be done before next.ServeHTTP is called on what will be the last request?

hmiddleware/connection_closer.go
```diff
@@ -0,0 +1,75 @@
...
+// ConnContext field.
+func ConnectionClosingMiddleware(mp metrics.Provider, maxRequests, cacheSize int) func(http.Handler) http.Handler {
+	var (
+		closed = mp.NewCounter("server.connection.closes.total")
+		mtx    sync.Mutex
+	)
+
+	cache, err := lru.New(cacheSize)
+	if err != nil {
+		log.Fatal(err)
```
danp: Should this use a passed-in logger or return an error if there's nothing else to log?

hmiddleware/connection_closer.go
iamjem: I guess I may not be 100% clear on the conventions for this connection close header. Is it expected that there'd be no other content to read from a response in the presence of the header? e.g. we should flip things to do this (i think this is what you're suggesting):

```
if shouldClose {
    closed.Add(1)
    w.Header().Add("connection", "close")
    return // <-- add this
}

// they're not at the limit yet, call next handler
next.ServeHTTP(w, r)

```

danp: If you can say "this will be the last request" I think it's fine to set the header and then call next.ServeHTTP. But if you call next.ServeHTTP first and then try and set the header I'm not sure it will work if w.WriteHeader has been called (or any body has been written). Maybe the http.Server bits will still see `Connection: close` and close the connection, I'm not sure. Using httptest.NewServer instead of a ResponseRecorder to get a more complete test rig might help build confidence.

iamjem: after poking at the net/http code this seems to be done cooperatively. whether the header is present in an http request from a client, or if its set in a response header, the desire to close the connection is respected in net/http. the idea here is once we hit the threshold we return the response like usual, but add an additional `connection: close` header.

hmiddleware/connection_closer.go
iamjem: good call, i was trying to avoid requiring a logger here, so i can just add the `error` to the return values.

iamjem:
excellent feedback @danp! after switching to an actual http server in the tests it became evident the order of adding the header does matter, makes total sense.

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/google/uuid v1.1.1
 	github.com/grpc-ecosystem/go-grpc-middleware v1.2.0
 	github.com/grpc-ecosystem/grpc-gateway v1.9.6
+	github.com/hashicorp/golang-lru v0.5.1
```
lxfontes: heads up this library uses MPL 2 license

hmiddleware/connection_closer.go
```diff
@@ -0,0 +1,76 @@
...
+	return context.WithValue(ctx, connCloseID, uuid.New().String())
+}
+
+// ConnectionClosingMiddleware returns middleware that will add a "Connection: close" responses header after a single
+// connection has produced maxRequests http requests. The cacheSize determines the size of the LRU cache which tracks
+// connection request counts. This middleware requires the http.Server sets ConnectionClosingContext for it's
+// ConnContext field.
+func ConnectionClosingMiddleware(mp metrics.Provider, maxRequests, cacheSize int) (func(http.Handler) http.Handler, error) {
+	var (
+		closed = mp.NewCounter("server.connection.closes.total")
```
lxfontes: the `server` is usually scoped by `METRICS_SUFFIX`
https://github.com/heroku/x/blob/96a1bfd8ec81b0ccb4966c6c8610d9e2ec7a1450/cmdutil/service/standard.go#L169

atm this creates `appname.server.server.connection.closes.total`

go.mod
iamjem: Is this an issue? looking at the FAQ it seems like this doesn't violate the license:
https://www.mozilla.org/en-US/MPL/2.0/FAQ/

From Q8 on the FAQ:
> You may distribute any executables you create under a license of your choosing, as long as that license does not interfere with the recipients' rights to the source under the terms of the MPL.

I guess we are doing some form of "distribution" here. But, the source is here for all to see, and we haven't modified the source of the original lru package with the MP license.

I did see another LRU package with an MIT license if folks are concerned at all though.

iamjem:
Some good feedback in sidebars, for the moment since this repo is pinned to go 1.12, and the changes I'm proposing involve a go 1.13 feature, we should hold off merging this until the repo is on 1.13 (which likely involves bumping other heroku repos to 1.13+ whom might be using heruoku/x).

Closing for now.

PR #134, opened 2020-05-06
Use Redigo 1.8 as it's the latest release

beanieboi:
Hey 👋 

i just discovered that the redigo versioning seems off. If you check the GitHub page https://github.com/gomodule/redigo/tags and https://github.com/gomodule/redigo/releases you can see that 1.8 is the latest release and that the 2.0 tag is several years old.

In numbers this is a "downgrade" but in reality this is now using a go modules compatible and more recent version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/go-kit/kit v0.9.0
 	github.com/gogo/protobuf v1.3.0 // indirect
 	github.com/golang/protobuf v1.3.2
-	github.com/gomodule/redigo v1.8.0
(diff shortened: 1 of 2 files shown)


danp:
Thanks! Does look like https://github.com/gomodule/redigo/releases/tag/v1.8.1 was just released as well which maybe includes a worthwhile fix.

beanieboi:
> Thanks! Does look like https://github.com/gomodule/redigo/releases/tag/v1.8.1 was just released as well which maybe includes a worthwhile fix.

cool, updated to 1.8.1

PR #135, opened 2020-05-07
logplex/encoding: Improve encoding performance

ypaq:
@freeformz was so kind to point out room for improvement here. Use `strings.Builder` instead of `fmt.Sprintf` to format logplex messages.

### DIFF
--- a/logplex/encoding/encoder.go
+++ b/logplex/encoding/encoder.go
@@ -107,17 +107,14 @@ func Encode(msg Message) ([]byte, error) {
 		return nil, errors.Wrap(ErrInvalidMessage, "version")
 	}
 
-	line := fmt.Sprintf("<%d>%d %s %s %s %s %s %s%s",
-		msg.Priority,
-		msg.Version,
-		msg.Timestamp.Format(SyslogTimeFormat),
-		stringOrNil(msg.Hostname),
-		stringOrNil(msg.Application),
-		stringOrNil(msg.Process),
-		stringOrNil(msg.ID),
-		sd,
-		msg.Message,
-	)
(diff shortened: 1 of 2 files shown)


cyx:
Nice!

freeformz:
This is great, but I'd like to point out the following:
```console
$ benchstat master ypaq-improve-logplex-encoder-perf      
name                old time/op    new time/op    delta
MessageToString-16     530ns ± 4%     342ns ± 2%  -35.52%  (p=0.000 n=10+10)

name                old alloc/op   new alloc/op   delta
MessageToString-16      160B ± 0%      128B ± 0%  -20.00%  (p=0.000 n=10+10)

name                old allocs/op  new allocs/op  delta
MessageToString-16      6.00 ± 0%      3.00 ± 0%  -50.00%  (p=0.000 n=10+10)
```

So time went down by 35%, bytes by 20%, and allocations by 50%.

But, have you considered just using a string?
https://github.com/heroku/x/compare/ypaq-improve-logplex-encoder-perf...freeformz:justAString?expand=1

Doing so leaves you with even better savings:
```console
$ benchstat master justAString
name                old time/op    new time/op    delta
MessageToString-16     530ns ± 4%     344ns ± 4%  -35.16%  (p=0.000 n=10+10)

name                old alloc/op   new alloc/op   delta
MessageToString-16      160B ± 0%       96B ± 0%  -40.00%  (p=0.000 n=10+10)

name                old allocs/op  new allocs/op  delta
MessageToString-16      6.00 ± 0%      2.00 ± 0%  -66.67%  (p=0.000 n=10+10)
```

ypaq:
I have not, thank you!

cyx:
#winning

PR #136, opened 2020-06-19
Creating Security Policy

brompwnie:
Adding security policy that is in alignment with other Salesforce repos.


brompwnie:
Thanks for the PR @danp 🎆

PR #139, opened 2020-07-06
cmdutil/service: support service.New(nil)

bernerdschaefer:
envdecode requires that at least one field in a provided config struct
match from the environment. This sometimes makes the standard service
awkward to work with, particularly when gradually migrating an existing
system.

It's common to see this while getting started:

```go
var cfg struct {
  AppName string `env:"APP_NAME"` // mirror APP_NAME from svclog for envdecode
}
s := service.New(&cfg)
```

This updates `service.New` to accept nil for for the appConfig argument
to ease conversion of services.

### DIFF
--- a/cmdutil/service/standard_test.go
+++ b/cmdutil/service/standard_test.go
@@ -1,0 +1,57 @@
+package service_test
+
+import (
+	"os"
+	"testing"
+	"time"
+
+	"github.com/heroku/x/cmdutil/service"
+)
+
+func TestNewNoConfig(t *testing.T) {
+	setupStandardConfig(t)
+
+	s := service.New(nil)
+
+	if s.Logger == nil {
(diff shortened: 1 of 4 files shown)


cmdutil/service/standard_test.go
```diff
@@ -0,0 +1,57 @@
...
+	if cfg.Val != time.Minute {
+		t.Fatalf("cfg.Val = %v want %v", cfg.Val, time.Minute)
+	}
+}
+
+func setupStandardConfig(t *testing.T) {
+	os.Setenv("APP_NAME", "test-app")
+	os.Setenv("DEPLOY", "test")
+
+	t.Cleanup(func() {
```
danp, line 53: Looks like something around CI needs to be bumped to go1.14 to use this.

PR #142, opened 2020-09-10
Add Helpers for 'Null Logger'

kennyp:
I've noticed a number of cases where a logger might not be given so null logger is used. I figured it would be helpful to centralize the pattern.

### DIFF
--- a/cmdutil/svclog/logger.go
+++ b/cmdutil/svclog/logger.go
@@ -2,6 +2,7 @@
 package svclog
 
 import (
+	"io/ioutil"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


vshiva (approved):
lgtm

cmdutil/svclog/logger.go
```diff
@@ -83,3 +84,19 @@ func (sl saramaLogger) Printf(format string, args ...interface{}) {
 	format = strings.TrimSpace(format)
 	sl.FieldLogger.Printf(format, args...)
 }
+
+// NewNullLogger returns a logger that discards the output useful for testing
```
bernerdschaefer, line 88: Since you mentioned "for testing" here, there is actually a package already with a null logger for testing: https://godoc.org/github.com/heroku/x/testing/testlog. It also includes useful assertions for testing structured log output.

kennyp:
Yeah, testing was one place it is being used. There are also places in the
code where providing a logger is optional and using a null logger makes it
easier for the rest of the code to be unaware that one wasn't provided.

I felt really weird pulling in a testing dependency in those cases.

PR #143, opened 2020-09-22
Add role validation to basicauth checking

slizco:

We'd like to extend the `basicauth` capabilities to include validation of role-based access. This [ADR](https://github.com/heroku/runtime/pull/4837) gives context.

## Description of changes

- Adds a `RoleValidator` interface that provides a `Validate(role, method string) bool` func that should return true when the `role` has access to `method`.
- Modifies `basicauth.GRPCAuthFunc` to call `roleValidator.Validate` if role validator is defined.
- Adds testing of the above.

## Runtime issue
https://trello.com/c/eEdSQMMa/193-implement-role-validator-in-heroku-x-basicauth

### DIFF
--- a/hmiddleware/basicauth/checker.go
+++ b/hmiddleware/basicauth/checker.go
@@ -50,7 +50,14 @@ func parseCredential(credential string) (Credential, error) {
 // Checker stores a set of valid credentials to be used when authenticating
 // HTTP requests using basic auth.
 type Checker struct {
-	credentials []Credential
+	credentials   []Credential
+	roleValidator RoleValidator
+}
+
+// RoleValidator providers Validate() which returns true
+// if a role has access to a particular request method.
+type RoleValidator interface {
+	Validate(role, method string) bool
 }
 
 // NewChecker returns a basic auth checker configured to use credentials
@@ -59,6 +66,11 @@ func NewChecker(credentials []Credential) *Checker {
 	return &Checker{
 		credentials: credentials,
 	}
+}
+
+// WithRoleValidator sets the roleValidator on a Checker.
+func (c *Checker) WithRoleValidator(rv RoleValidator) {
+	c.roleValidator = rv
 }
 
 // Valid is true if username and password represent acceptable credentials.
--- a/hmiddleware/basicauth/grpc.go
+++ b/hmiddleware/basicauth/grpc.go
@@ -44,7 +44,10 @@ func GRPCAuthFunc(checker *Checker) func(ctx context.Context) (context.Context, 
 			}
 
 			if !checker.roleValidator.Validate(user, method) {
-				return nil, grpc.Errorf(codes.PermissionDenied, fmt.Sprintf("permission denied. user \"%s\" does not have access to \"%s\"", user, method)) //nolint:staticcheck
+				return nil, grpc.Errorf( //nolint:staticcheck
+					codes.PermissionDenied,
+					fmt.Sprintf("permission denied. user \"%s\" does not have access to \"%s\"", user, method),
+				)
 			}
 		}
 
--- a/hmiddleware/basicauth/grpc.go
+++ b/hmiddleware/basicauth/grpc.go
@@ -3,6 +3,7 @@ package basicauth
 import (
 	"context"
 	"encoding/base64"
+	"fmt"
 	"strings"
 
 	grpc_auth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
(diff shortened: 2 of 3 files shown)


danp:
This maybe trends from authentication to authorization, and I'd say so far basicauth has only been about authentication. Should authorization happen in a more distinct place to keep it separate from authentication?

slizco:
@danp Yes, you're right, it is more authorization. We'd like to use the grpc interceptor model to accomplish this though, and this is the simplest way for us to do so. We could write a a whole separate interceptor if need be though. I do think adding the role validation as an optional functionality here is okay. I'll let others weigh in too

hmiddleware/basicauth/checker.go
```diff
@@ -50,7 +50,14 @@ func parseCredential(credential string) (Credential, error) {
 // Checker stores a set of valid credentials to be used when authenticating
 // HTTP requests using basic auth.
 type Checker struct {
-	credentials []Credential
+	credentials   []Credential
+	roleValidator RoleValidator
```
galbacarys, line 54: Is it possible we could want more than one role validator?

slizco: Mm I can't see a reason we would. The Credentials field is a list because they are pairs of username/passwords, where each client gets its own pair.

galbacarys (approved):
LGTM pending ci passing. I think the conflation of AuthN/AuthZ isn't the biggest deal in the world since it's such a simple scheme and the authZ portion is optional.

slizco:
Yeah the failure is due to test coverage I believe. I'm going to add a couple more cases

hmiddleware/basicauth/grpc.go
```diff
@@ -34,6 +35,22 @@ func GRPCAuthFunc(checker *Checker) func(ctx context.Context) (context.Context,
 			return nil, grpc.Errorf(codes.PermissionDenied, "permission denied") //nolint:staticcheck
 		}
 
+		if checker.roleValidator != nil {
+			// According to grpc docs, this method will be in the form:
+			// "/service/method".
+			method, ok := grpc.Method(ctx)
+			if !ok {
+				return nil, grpc.Errorf(codes.PermissionDenied, "permission denied. no grpc method found on context") //nolint:staticcheck
```
ypaq, line 43: Is the error message relayed back to the client?

ypaq (approved):
Looks good to me, thank you.

ypaq:
To @danp's point - I agree that they are different concerns that we might want to consider to separate. Alternatively to the implementation here, the grpc_middleware supports chaining of interceptors which we could utilize to separate the concerns of authn and authz. We could leverage the context to pass the role. If that would be a better implementation I'm not sure. The fact that we're using basic auth is a tirefire on its own, and this change definitely doesn't make it worse.

borncrusader (approved):
LGTM. Great work, @slizco

hmiddleware/basicauth/grpc.go
```diff
@@ -34,6 +35,22 @@ func GRPCAuthFunc(checker *Checker) func(ctx context.Context) (context.Context,
...
+			}
+
+			if !checker.roleValidator.Validate(user, method) {
+				return nil, grpc.Errorf( //nolint:staticcheck
+					codes.PermissionDenied,
+					fmt.Sprintf("permission denied. user \"%s\" does not have access to \"%s\"", user, method),
```
iamjem, line 49: you can use `%q` instead of escaping the quotes here

iamjem (commented):
I have similar feels as @danp, its common for authentication and authorization to exist in separate layers/phases of request processing (the order matters of course!). So the authz part would just be a separate unary/stream interceptor, and then the checker type could be left as-is. There's already some utilities for collecting grpc options--including interceptors--and we could just have this be another optional interceptor to be provided and included in the chain.

I also don't think we need to put everything that feels general purpose in heroku/x immediately, we can keep it local to a project while we iterate and polish the idea, and move it here when it makes sense (for example when multiple projects need access to said functionality).

slizco:
We do have this same code in `heroku/runtime/lib`, so I'm okay with expanding upon that for now. I'm going to work on writing this as a separate interceptor over there and see how that looks

hmiddleware/basicauth/grpc.go
slizco: Mmm good question. Don't know, but I can check what we do

slizco:
Closing this stale PR. The role-based authorization functionality has been added to `heroku/runtime/lib` instead.

PR #145, opened 2021-03-11
Add ECDSA and CHACHA, remove CBC

fernomac:
Adding ECDSA ciphers to the SF ciphers list to account for customers with EC certs. Per conversations with security, adding CHACHA variants and removing CBC variants while we're here. (And fixing an unrelated compiler warning re: casting `x509.PublicKeyAlgorithm` to a string instead of using its String func.)


slizco (approved):
Do you mind updating this doc too? :)
https://salesforce.quip.com/1ksHAZKkNYSA

ypaq (approved):
💯

borncrusader (commented):
LGTM!

PR #150, opened 2021-09-27
Add utils for setting up lambda handlers

slizco:
This addition includes a utility package that can be used to standardize the configuration and handler set-up of our Lambda functions. It is modeled off of the `cmdutil.Service` type.

The go-kit `metrics.Provider` interface has been extended to include a `Flush()` method that operates much the same as `Stop()`, but does not tear down the provider and may be called multiple times. In some cases, `Flush()` is a no-op. It has been added here so that recorded metric values may be optionally flushed to the provider service from within the handler. This has proved necessary when using the Librato & OTEL providers, to ensure data are not lost when the execution context of the Lambda invocation is torn down.

### DIFF
--- a/lambda/function.go
+++ b/lambda/function.go
@@ -7,6 +7,7 @@ import (
 	"github.com/joeshaw/envdecode"
 	"github.com/oklog/run"
 	"github.com/sirupsen/logrus"
+	"go.opentelemetry.io/otel/attribute"
 
 	"github.com/heroku/x/cmdutil"
 	"github.com/heroku/x/cmdutil/metrics"
@@ -16,6 +17,12 @@ import (
 	kitmetrics "github.com/heroku/x/go-kit/metrics"
 	"github.com/heroku/x/go-kit/metrics/l2met"
 	"github.com/heroku/x/go-kit/metrics/multiprovider"
+	xotel "github.com/heroku/x/go-kit/metrics/provider/otel"
+)
+
+const (
+	// This is the key for the "function" OTEL attribute. The value should be the function name.
+	functionKey = "function"
 )
 
 // Function defines configuration of a Lambda function.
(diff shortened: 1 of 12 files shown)


voidlock:
This change will force us down a path of further deviation from the [go-kit standard Provider interface](https://pkg.go.dev/github.com/go-kit/kit@v0.12.0/metrics/provider#Provider). Could we accomplish this enhancement using composable interfaces somehow?

ypaq (approved):
Looks good to me! Thank you 🎉

slizco:
@voidlock I think that's a good point about deviating more from the standard `Provider` interface. We have _already_ deviated so I would question if it matters much at this point. The `Flush` will for most providers be a no-op. As for introducing a composable interface, something like `FlushableProvider` could work, yes. I'd be okay with that as long as we actually add a `Flush()` method to each of the satisfying provider types because I want to be able to call it without casting in the lambda handler code.

We could alternatively remove the `MetricsProvider` field from `lambda.Function` and replace it within a method, `MetricsProvider()` that returns a new metrics provider. That would allow us to create, start and stop the metrics provider all within the passed `handler`. The explicit calling of `Stop()` would ensure the metric values are flushed.

Opinions on this? @ypaq @borncrusader @agnaite @fernomac

lambda/function.go
```diff
@@ -0,0 +1,152 @@
...
+	// Logger is a field logger.
+	Logger logrus.FieldLogger
+	// Metrics provider defines interactions for recording metrics.
+	MetricsProvider kitmetrics.Provider
+
+	g run.Group
+}
+
+// New creates a new Function with a configured logger, rollbar agent and metrics provider.
+func New(config interface{}) *Function {
```
borncrusader, line 50: Nice! :)

borncrusader (approved):
LGTM

PR #151, opened 2021-10-01
Update otel provider

ypaq:
* Update otel dependency
* Include provider attributes in metrics


agnaite (commented):
curious if you tested this to make sure it works? I’m a little worried we are going to see some attributes twice (the global provider attributes + the metric attributes), but maybe it removes duplicates 🤔

ypaq:
I vendor hacked this to see if it works which it did. Not 100% sure if there are dups, but that's a good point. Waiting to hear back if it's an issue on the other side. We might not need this.

slizco (approved):
lgtm

PR #156, opened 2021-11-17
Add MustProvider() for otel provider set-up, fix service.namespace attr

slizco:
To avoid repeating the definition of this `MustProvider` func across our various codebases, we add it here. This should reduce redundant code and help us standardize on attribute settings.

We also update the value of `service.namespace` from a hard-coded value of `heroku` to a dynamic value passed upon provider set-up. This will allow us to set a value that varies per deploy/cloud, which is more in-line with OpenTelemetry standards: https://opentelemetry.lightstep.com/core-concepts/attributes-and-labels/


cmdutil/metrics/otel/otel.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/heroku/x/cmdutil/metrics/honeycomb"
+	"github.com/heroku/x/go-kit/metrics"
+	otelprovider "github.com/heroku/x/go-kit/metrics/provider/otel"
+)
+
+// Config is a reusable configuration struct that contains the necessary
+// environment variables to setup an metrics.Provider
+type Config struct {
+	Enabled             bool     `env:"ENABLE__COLLECTION"`
```
cgfuh, line 21: silly comment here, but, maybe it's a typo. Why the double underscore?

voidlock (approved):
Thanks for adding this. The only thing I spotted was what @cgfuh already pointed out with the double underscore.

cmdutil/metrics/otel/otel.go
slizco: Nice catch! This is a mistake

PR #158, opened 2021-11-17
Include the otel config as part of general metrics config

slizco:
This puts the `otel.Config` as a field on the general `metrics.Config`.


cgfuh (approved):
lgtm

PR #159, opened 2021-11-17
Add support for Heroku Redis 6 using TLS

voidlock:
Connect to any Heroku Redis plan using Redis 6 (default) require TLS (see https://devcenter.heroku.com/articles/heroku-redis#security-and-compliance).

This commit follows the directions outlined in [Connecting to Go](https://devcenter.heroku.com/articles/heroku-redis#connecting-in-go) to allow for both TLS and non-TLS connections.

### DIFF
--- a/cmdutil/redispool/redispool.go
+++ b/cmdutil/redispool/redispool.go
@@ -70,7 +70,10 @@ func (c Config) Pools() ([]*redis.Pool, error) {
 func newPool(cfg Config) *redis.Pool {
 	return &redis.Pool{
 		Dial: func() (redis.Conn, error) {
-			conn, err := redis.DialURL(cfg.URL)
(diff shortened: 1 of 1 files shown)


borncrusader (approved):

slizco:
@voidlock Is the connection secured in some way? I see the docs talk about using the stunnel feature: https://devcenter.heroku.com/articles/securing-heroku-redis#stunnel-overview

voidlock:
> Is the connection secured in some way? I see the docs talk about using the stunnel feature: https://devcenter.heroku.com/articles/securing-heroku-redis#stunnel-overview

The connection is "encrypted". It doesn't seem to matter if you use the built-in TLS support from Redis 6 or the stunnel feature. Heroku Redis uses a self-signed certificate in both cases. It also seems the stunnel feature is only supported on the older Redis 4 versions.

PR #161, opened 2021-11-19
adds ability to choose the hash used for sharding

voidlock:
This change adds complete control over the hashing algorithm used by the
Sharder. Sometimes a user may want to tightly control allocations, other
times a user may just want pure concurrency.

The existing behavior remains unchanged so that existing users of the
library are not surprised. This existing behavior is now wrapped as the
`WithLockingHasher` option.

A second option, `WithLockFreeHasher` is available that uses the
same fnv hashing algorithm without the use of locks.

A third option, `WithHasher`, allows for complete customization.

### DIFF
--- a/sharder/sharder.go
+++ b/sharder/sharder.go
@@ -6,35 +6,94 @@ import (
 	"sync"
 )
 
-// New returns a new Sharder with the specified number of shards.
-func New(total int) *Sharder {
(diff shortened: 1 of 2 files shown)


wchrisjohnson (approved):

borncrusader:
LGTM

PR #163, opened 2021-12-16
Fix typo in default metrics destinations

ypaq:


borncrusader (approved):
LGTM

slizco:
@ypaq I actually think this is right the way it is: https://go.dev/play/p/ptVxUXWxY4D

slizco:
@ypaq Never mind, I had this change backwards in my head. Good catch!

slizco (approved):
awesome

PR #165, opened 2022-01-06
Collect OTEL metric data once a min

slizco:

Some of our alerting for our Runtime services will rely on percentage-based metrics computed using other simple metrics. Each percentage-based metric is computed over a single event, so numbers can fluctuate widely when the underlying metrics are `undefined`. Collecting the metrics at a lower frequency (less often) will ensure more of the underlying metrics are defined on the events, making the derived column value more accurate and meaningful. 

## Description of changes

- Hard-code the `CollectPeriod` to be `1m` on the OTEL controller.

### DIFF
--- a/go-kit/metrics/provider/otel/provider.go
+++ b/go-kit/metrics/provider/otel/provider.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


ypaq (approved):
LGTM

PR #166, opened 2022-01-10
Pass collector URL by value

wchrisjohnson:
[Buzz](https://github.com/heroku/buzz) needs to be able to configure multiple providers multiple times as a part of its operational approach. The existing code works the first time a provider is initiated, but if it is attempted a second time, it fails. 

The `collectorURL` is passed in by reference, and the code inside the method mutates the URL; no bueno. This change passes the URL by value, so the mutation has no base side effects.


wchrisjohnson:
@slizco do we need a test to cover this case?

slizco (approved):
nice!

PR #167, opened 2022-01-14
Fix otel.NewCardinalityCounter

tbalthazar:
The `otel.NewCardinalityCounter` method returns
`&xmetrics.HLLCounter{}`, which means `HLLCounter.counter` is `nil`,
which will prevent code from calling `.Insert()` on it without panicing.

Use the `NewHLLCounter` method insead, it correctly initializes the
`counter` field.


wchrisjohnson (approved):
LGTM!

PR #168, opened 2022-01-14
Provide opt to config OTEL provider collect period

slizco:
This will remove the hard-coding of `CollectPeriod = 1m` and allow for the caller to configure the period via an option to `xotel.New(...)`.


tbalthazar (approved):
🚀

PR #169, opened 2022-01-26
[WIP] Use `HistogramDistribution` as default aggregator

voidlock:
The are two problems with the existing `ExactDistribution` aggregator:

1. It's gone in go.opentelemetry.io/otel/sdk/metric@v0.26.0/
2. It ships every metrics ever collected causing huge payloads for busy
   services.

This commit resolves that by switching the HistogramDistribution, which
will send much smaller payloads. It is most similar to the original
histogram implementation found in https://pkg.go.dev/github.com/heroku/x@v0.0.47/go-kit/metrics/provider/librato#Histogram.

## Checklists

- [ ] Find way to adjust histogram buckets

### DIFF
--- a/cmdutil/metrics/otel/config.go
+++ b/cmdutil/metrics/otel/config.go
@@ -11,6 +11,7 @@ import (
 type Config struct {
 	Enabled             bool     `env:"ENABLE_OTEL_COLLECTION"`
 	CollectorURL        *url.URL `env:"OTEL_COLLECTOR_URL"`
+	UseExactAggregator  bool     `env:"OTEL_USE_EXACT_AGGREGATOR"`
 	MetricsDestinations []string `env:"OTEL_METRICS_DESTINATIONS,default=honeycomb;argus"`
 	Honeycomb           honeycomb.Config
 }
--- a/cmdutil/metrics/otel/config.go
+++ b/cmdutil/metrics/otel/config.go
@@ -9,9 +9,9 @@ import (
 // Config is a reusable configuration struct that contains the necessary
 // environment variables to setup an metrics.Provider
 type Config struct {
-	Enabled             bool     `env:"ENABLE_OTEL_COLLECTION"`
-	CollectorURL        *url.URL `env:"OTEL_COLLECTOR_URL"`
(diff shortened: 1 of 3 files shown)


agnaite (approved):
LGTM

cmdutil/metrics/otel/config.go
```diff
@@ -9,8 +9,9 @@ import (
 // Config is a reusable configuration struct that contains the necessary
 // environment variables to setup an metrics.Provider
 type Config struct {
-	Enabled             bool     `env:"ENABLE_OTEL_COLLECTION"`
-	CollectorURL        *url.URL `env:"OTEL_COLLECTOR_URL"`
 	MetricsDestinations []string `env:"OTEL_METRICS_DESTINATIONS,default=honeycomb;argus"`
 	Honeycomb           honeycomb.Config
+	CollectorURL        *url.URL `env:"OTEL_COLLECTOR_URL"`
+	Enabled             bool     `env:"ENABLE_OTEL_COLLECTION"`
+	UseExactAggregator  bool     `env:"OTEL_USE_EXACT_AGGREGATOR"`
```
ypaq, line 16: There are probably several options for aggregation. If we have to utilize a different one in another case we would end up with competing boolean config options. My proposal would be to map the config to an aggregator which allows you to leave it unset with the default option:

```
Aggregator string `env:"OTEL_AGGREGATOR"`
```

voidlock: That is how I tested it elsewhere. 

The new default is a good default choice for most metrics exports (see: https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/selector/simple#NewWithHistogramDistribution) which seems to make sense for a library this one.

The presence of the config var was mostly to allow folks to option of immediately turning the new behaviour on/off while addressing any changes that will occur with the exported metrics.

If anyone absolutely needs to override this default behaviour and use the other aggregator it is easy enough to do by using  `otel.WithAggregator(simple.NewWithInexpensiveDistribution())` to construct their Otel provider.

ypaq (approved):

voidlock:
Closing this in favour of #170

PR #170, opened 2022-01-31
Workaround otel histogram options

voidlock:
The commit works around the disconnect between the OTEL metrics Meter interface for creating histograms and the actual Aggregator used to collect and report on the data. Presently, there is no way to pass bucket definitions to the aggregator except globally for all histograms.

This commit resolves that by allowing for the use of a custom `AggregatorSelector` that caches any options and looks them up when a new `Aggregator` is requested by the OTEL controller.

At this point in time, I am leaving everything but the OTEL provider implementation alone. If you use use the `ExplicitHistogram` it will simply create an older style Histogram with a bucket count set to the length of the boundaries array.


slizco (approved):
I think this makes sense. I wonder if we may also want to provider an exported `DefaultBoundaries` and/or use those boundaries for `NewHistogram`. As you mentioned, most histograms are measuring a distribution of request latencies, so they're likely to be in the range of 0-30s, with most data points falling under 1s. We can add that later though if we find ourselves repeating boundary definitions.

Is there anything preventing us from merging this soon? I believe it does not conflict with existing histogram functionality.

wchrisjohnson (approved):
LGTM on my initial pass - would be good to get this into the wild and see how well it works.

PR #171, opened 2022-02-09
add new highcardmetrics pkg to grpcmetrics

agnaite:
## Context

We want to add two new options to the grpc server, which allows us to collect metrics using attributes. 

The new `highcardmetrics` pkg provides high cardinality unary and stream server interceptors. The interceptors take an `explicit` bool arg, which makes the histogram type configurable. Passing in `true` will create explicit histograms with a default 5s distribution. 

## Usage

To initialize a grpc server with high cardinality unary and stream servers using explicit histograms:

```go
gSrv, hSrv := grpcserver.NewStandardH2C(
  apiSrv,
  // high cardinality interceptors using explicit histograms
  grpcserver.HighCardInterceptors(
     highcardmetrics.NewUnaryServerInterceptor(o.metricsProvider, true), 
     highcardmetrics.NewStreamServerInterceptor(o.metricsProvider, true),
   ),
)
```

### DIFF
--- a/grpc/grpcmetrics/highcardmetrics/server.go
+++ b/grpc/grpcmetrics/highcardmetrics/server.go
@@ -158,5 +158,5 @@ func isCanceled(err error) bool {
 }
 
 func ms(d time.Duration) float64 {
-	return float64(d) / float64(time.Millisecond)
+	return float64(d.Milliseconds())
 }
--- a/grpc/grpcmetrics/highcardmetrics/server.go
+++ b/grpc/grpcmetrics/highcardmetrics/server.go
@@ -75,7 +75,7 @@ type serverStream struct {
 	explicitHistograms bool
 }
 
-//// RecvMsg implements the grpc.Stream interface.
+// RecvMsg implements the grpc.Stream interface.
 func (ss *serverStream) SendMsg(m interface{}) (err error) {
 	defer func(begin time.Time) {
 		instrumentStreamSend(ss.reg, ss.explicitHistograms, ss.labels, time.Since(begin), err)
@@ -84,7 +84,7 @@ func (ss *serverStream) SendMsg(m interface{}) (err error) {
 	return ss.ServerStream.SendMsg(m)
 }
 
-//// RecvMsg implements the grpc.Stream interface.
+// RecvMsg implements the grpc.Stream interface.
 func (ss *serverStream) RecvMsg(m interface{}) (err error) {
 	defer func(begin time.Time) {
 		instrumentStreamRecv(ss.reg, ss.explicitHistograms, ss.labels, time.Since(begin), err)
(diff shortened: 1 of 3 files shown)


grpc/grpcmetrics/highcardmetrics/server.go
```diff
@@ -0,0 +1,162 @@
...
+// NewStreamServerInterceptor returns an interceptor for stream server calls
+// which will report metrics to the given provider.
+func NewStreamServerInterceptor(p metrics.Provider) grpc.StreamServerInterceptor {
+	r0 := metricsregistry.New(p)
+	return func(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) (err error) {
+		r1 := metricsregistry.NewPrefixed(r0, "grpc.server")
+
+		service, method := parseFullMethod(info.FullMethod)
+
+		labels := []string{"service", service, "method", method}
```
voidlock, line 45: Will this `service` conflict (or be confused) with the existing `service*` columns?

grpc/grpcmetrics/highcardmetrics/server.go
```diff
@@ -0,0 +1,162 @@
...
+	if st, ok := status.FromError(err); ok {
+		return st.Code() == codes.Canceled
+	}
+
+	return false
+}
+
+func ms(d time.Duration) float64 {
+	return float64(d) / float64(time.Millisecond)
+}
```
voidlock, line 162: There is a much easier way to do this now https://pkg.go.dev/time#Duration.Milliseconds

voidlock (commented):
Yup, this is exactly what I was thinking.

grpc/grpcmetrics/highcardmetrics/server.go
agnaite: very true. how do you feel about `grpc.service` and `grpc.method`? Or any other suggestions welcome!

voidlock: It's like you can read my mind, I love those choices! 

In fact, I was wondering about changing ours in the httpmetrics middleware to use `http.method`, `http.path`, `http.response-status`.

grpc/grpcmetrics/highcardmetrics/server.go
```diff
@@ -0,0 +1,162 @@
...
+// NewUnaryServerInterceptor returns an interceptor for unary server calls
+// which will report metrics to the given provider.
+func NewUnaryServerInterceptor(p metrics.Provider) grpc.UnaryServerInterceptor {
+	r0 := metricsregistry.New(p)
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (_ interface{}, err error) {
+		r1 := metricsregistry.NewPrefixed(r0, "grpc.server")
+
+		defer func(begin time.Time) {
+			service, method := parseFullMethod(info.FullMethod)
+			labels := []string{"service", service, "method", method, "response-status", code(err)}
```
voidlock, line 27: These labels will need to get tweaked as well.

voidlock (approved):
Looks great!

PR #174, opened 2022-05-03
Improve grpc basicauth errors

ypaq:


hmiddleware/basicauth/grpc.go
```diff
@@ -25,13 +25,11 @@ func GRPCAuthFunc(checker *Checker) func(ctx context.Context) (context.Context,
 
 		user, pass, ok := parseBasicAuth(blob)
 		if !ok {
-			//TODO: SA1019: grpc.Errorf is deprecated: use status.Errorf instead.  (staticcheck)
-			return nil, grpc.Errorf(codes.Unauthenticated, "unauthenticated") //nolint:staticcheck
+			return nil, status.Errorf(codes.Unauthenticated, "unauthenticated")
```
agnaite, line 28: looking at the docs i feel like this should be a `status.Error` vs `status.Errorf`? https://pkg.go.dev/google.golang.org/grpc/status#Error

agnaite (approved):
LGTM, 1 nit, see comment :)

hmiddleware/basicauth/grpc.go
ypaq: Agreed!

PR #176, opened 2022-07-07
Support dyno ID verification

tt:
This is a counterpart for heroku/dogwood#2762 et al. to support dyno ID verification in Go components.

It includes a package responsible for verification and another to ease testing of the functionality.


tbalthazar (approved):
🚀

PR #177, opened 2022-07-20
Quick enhancements to redispool and available histogram distributions

voidlock:
A couple of quick enhancements.

- Updated redigo dependency.
- Converted the `redispool` wrapper to use `DialContext` over `Dial` to allow for proper propagation of contexts that may have deadlines or otherwise be cancelled.
- Added a TenSecondDistribution to provide a middle between the 5s and 30s distributions for time based histograms.

### DIFF
--- a/cmdutil/redispool/redispool.go
+++ b/cmdutil/redispool/redispool.go
@@ -3,6 +3,7 @@
 package redispool
 
 import (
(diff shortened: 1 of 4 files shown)


agnaite (approved):
LGTM!

PR #178, opened 2022-07-27
deal with Redis 4, 5 & 6 compatibility issues

voidlock:
For Heroku Redis, we need to strip the username from the URL. After the release of Redis 6, clients began to support the new AUTH command that uses 2 arguments. Heroku Redis removed the `h` username from connection URLs, but only on new addons. Existing addons do not have the `h` removed so we must check for and remove it here.

See https://devcenter.heroku.com/changelog-items/1932

### DIFF
--- a/cmdutil/redispool/redispool.go
+++ b/cmdutil/redispool/redispool.go
@@ -4,6 +4,7 @@ package redispool
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


agnaite (approved):
LGTM

PR #180, opened 2022-10-17
allow redigo pools to wait

voidlock:
Calls to redigo's `pool.Get()` will only block when configured to do so. The default behaviour will simply return instead of waiting (see https://github.com/gomodule/redigo/blob/master/redis/pool.go#L154-L156).

This PR introduces extra configuration to allow customization of this behaviour.

### DIFF
--- a/cmdutil/redispool/redispool.go
+++ b/cmdutil/redispool/redispool.go
@@ -41,6 +41,10 @@ type Config struct {
 	// When set to zero, aliveness is not checked.
 	IdleProbeInterval time.Duration `env:"REDIS_IDLE_PROBE_INTERVAL,default=0s"`
 
+	// Wait when true and the pool is at the MaxActiveConns limit, calls to Get()
+	// wait for a connection to be returned to the pool before returning.
+	Wait bool `env:"REDIS_WAIT,default=false"`
+
 	// AttachmentNames is an optional configuration which allows to set up
 	// connection pools to each redis attachment using the Pools() function. The
 	// same configuration parameters apply for each pool, except the URL which is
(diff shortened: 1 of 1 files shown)


cmdutil/redispool/redispool.go
```diff
@@ -41,6 +41,10 @@ type Config struct {
 	// When set to zero, aliveness is not checked.
 	IdleProbeInterval time.Duration `env:"REDIS_IDLE_PROBE_INTERVAL,default=0s"`
 
+	// Wait when true and the pool is at the MaxActiveConns limit, calls to Get()
+	// wait for a connection to be returned to the pool before returning.
+	Wait bool `env:"REDIS_WAIT,default=false"`
```
voidlock, line 46: NOTE: This is the default behaviour when you don't explicitly set it.

fernomac: Is there a timeout on how long we'll wait?

(Which is to say: do we think this is a case of a temporary spike in load and things will calm down if we wait a bit, or do we think it's connections getting leaked such that waiting will just show up as errors/timeouts later on?)

ypaq: I would guess that if we use [pool.GetContext](https://github.com/gomodule/redigo/blob/master/redis/pool.go#L198) on the consumer side we'd be ok.

voidlock: logtails is using `pool.GetContext` already in the hot path, `/drains`. When I update https://github.com/heroku/go-logplex/pull/327 to use this instead of forcing it I can update the `/stream` (aka subscribers) to use `pool.GetContext` as well.

PR #181, opened 2022-10-20
Add low card provider for optionally skipping labels

slizco:
This PR adds a `lowcard` package that may be used to wrap `metrics.Provider`s. 

Labels added to instruments using `With()` will be ignored unless explicitly marked for inclusion. Label inclusion is indicated by passing the list to `NewLowCardinalityWrappedProvider()` as an argument when creating the low card wrapper.


agnaite (approved):
Thank you!

PR #182, opened 2022-10-26
Add clock pkg for simple time abstractions

slizco:
The `clock` pkg provides useful abstraction around Go's std time telling/measuring package.

Instead of direct calls to `time.Now()`, you would construct a struct with a `clock.Clock` field:
```
var a =  struct {
  Clock clock.Clock
}{}

a.Clock = clock.Default
```
Now you may retrieve current time with:
```
a.Clock.Now()
```

In unit tests, mocking out your clock is simple with use of the `clocktest` pkg:
```
t0 := time.Now()
t1 := time.Now().Add(time.Second)

a.Clock = clocktest.New(t0, t1)

a.Clock.Now() // equal to t0
a.Clock.Now() // equal to t1
```

### DIFF
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -1,0 +1,17 @@
+package clock
+
+import "time"
+
+// Default is the default clock.
+var Default = Func(time.Now)
+
+// Clock tells you the current time.
+type Clock interface {
+	Now() time.Time
+}
+
+// Func is a function that returns a time.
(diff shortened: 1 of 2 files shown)


slizco:
Closing this since this repo already pulls in [benbjohnson/clock](https://github.com/benbjohnson/clock).

PR #179, opened 2022-09-22
Update deps

ricardochimal:
Updating dependencies & some misc cleanup work

- [ ] Update to a more recent go version
- [ ] Remove librato provider support
- [ ] Migrate to github actions
  - [ ] unit tests
  - [ ] code coverage step
  - [ ] linting
  - [ ] validating gomod
- [ ] Makefile cleanup

PR #193, opened 2023-04-07
Only start l2met provider when OTEL not enabled

slizco:
We don't want to run an `l2met` provider by default. With a [prior change](https://github.com/heroku/x/pull/179/files#diff-9692d64d304dc9da6aaa7ad1bac87402306ed56bb5507e9665598bc9e0f7b997L76-R77), that is now the behavior.

This PR changes the default behavior so that an `l2met` provider will only be started if OTEL is not enabled. This behavior can be overridden by setting the `METRICS_ENABLE_L2MET_OVERRIDE` env var, in which case, an `l2met` provider will be started regardless of the OTEL settings.

### DIFF
--- a/cmdutil/metrics/metrics.go
+++ b/cmdutil/metrics/metrics.go
@@ -13,5 +13,8 @@ type Config struct {
 	Source         string        `env:"METRICS_SOURCE"`
 	Prefix         string        `env:"METRICS_PREFIX"`
 	DefaultTags    []string      `env:"METRICS_DEFAULT_TAGS"`
-	OTEL           otel.Config
+	// If OTEL is enabled, l2met is disabled, by default.
+	// Setting this value to `true` overrides that default.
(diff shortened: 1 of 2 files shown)


cmdutil/metrics/metrics.go
```diff
@@ -13,5 +13,8 @@ type Config struct {
 	Source         string        `env:"METRICS_SOURCE"`
 	Prefix         string        `env:"METRICS_PREFIX"`
 	DefaultTags    []string      `env:"METRICS_DEFAULT_TAGS"`
-	OTEL           otel.Config
+	// If OTEL is enabled, l2met is disabled, by default.
+	// Setting this value to `true` overrides that default.
+	L2MetEnabled bool `env:"METRICS_ENABLE_L2MET_COLLECTION"`
```
ricardochimal, line 18: It doesn't seem apparent that this is to override behavior

maybe `METRICS_ENABLE_L2MET_OVERRIDE` ?

ricardochimal (approved):

PR #194, opened 2023-05-03
Check if MetricsProvider is nil, avoid panics

slizco:
When `service.Run()` exits, it's possible to panic because `MetricsProvider` may be `nil`.

In the future, we should perhaps set-up an OTEL provider or a NOOP provider by default.


ypaq (approved):
Should we also check if the logger is nil?

slizco:
@ypaq I think it can't be `nil`. `service.New` will fail if so

PR #195, opened 2023-05-10
cmdutil/service: improve panic reporting

ypaq:
## Rationale

We'd like to see a metric when we see panics in a standard service.

## Changes
* Add `ReportPanic` function to `cmdutil/metrics` that creates a counter `panic` on the metrics provider configured with a service.
* Move `ReportPanic` logging function to `cmdutil/svclog`
* L2met metrics provider now logs on `Flush`
* Add integration test that shows logs on a service panic for l2met and standard logger

## Meta 
* [W-13145387](https://gus.lightning.force.com/a07EE00001QhRMcYAN)

### DIFF
--- a/cmdutil/service/integration_test.go
+++ b/cmdutil/service/integration_test.go
@@ -1,0 +1,42 @@
+//go:build integration
+// +build integration
+
+package service
+
+import (
+	"os"
+	"testing"
+
+	"github.com/heroku/x/cmdutil"
+	"github.com/heroku/x/go-kit/metrics/l2met"
+)
+
+func TestPanicReporting(t *testing.T) {
+
+	os.Setenv("APP_NAME", "test-app")
+	os.Setenv("DEPLOY", "test")
+
+	t.Cleanup(func() {
+		os.Unsetenv("APP_NAME")
+		os.Unsetenv("DEPLOY")
+	})
+
+	var cfg struct {
+		Val string `env:"TEST_VAL,default=test"`
+	}
+
+	s := New(&cfg)
+
+	l2met := l2met.New(s.Logger)
+	s.MetricsProvider = l2met
+	s.Add(cmdutil.NewContextServer(l2met.Run))
+
+	f := func() error {
+		panic("test panic")
+		return nil
+	}
+
+	s.Add(cmdutil.ServerFunc(f))
+	s.Run()
+
+}
--- a/cmdutil/service/standard.go
+++ b/cmdutil/service/standard.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/heroku/x/cmdutil"
 	"github.com/heroku/x/cmdutil/debug"
+	"github.com/heroku/x/cmdutil/metrics"
 	"github.com/heroku/x/cmdutil/oc"
 	"github.com/heroku/x/cmdutil/rollbar"
 	"github.com/heroku/x/cmdutil/signals"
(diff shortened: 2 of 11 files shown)


ypaq:
```
▸ go test -tags integration ./cmdutil/service/...
time="2023-05-10T14:38:07-07:00" level=info app=test-app at=skipping-rollbar deploy=test
time="2023-05-10T14:38:07-07:00" level=panic msg="test panic" app=test-app deploy=test
time="2023-05-10T14:38:07-07:00" level=info app=test-app at=metrics count#panic=1 deploy=test
panic: test panic [recovered]
        panic: (*logrus.Entry) 0x140000beaf0 [recovered]
        panic: (*logrus.Entry) 0x140000beaf0
[...]
```

cmdutil/service/integration_test.go
```diff
@@ -0,0 +1,42 @@
...
+	s.Add(cmdutil.NewContextServer(l2met.Run))
+
+	f := func() error {
+		panic("test panic")
+		return nil
+	}
+
+	s.Add(cmdutil.ServerFunc(f))
+	s.Run()
+
```
slizco, line 41: Is it intentional that this integration test have no assertions? Is the expectation that you would just check logs for the panic info?

cmdutil/service/standard.go
```diff
@@ -101,7 +102,13 @@ func New(appConfig interface{}, ofs ...OptionFunc) *Standard {
 // Add adds cmdutil.Servers to be managed.
 func (s *Standard) Add(svs ...cmdutil.Server) {
 	for _, sv := range svs {
-		s.g.Add(sv.Run, sv.Stop)
+		sv := sv
+		runWithPanicReport := func() error {
```
slizco, line 106: Why not do this wrapping in `New()` by utilizing the `options`? That feels a little more in-line with the existing options pattern.

go-kit/metrics/l2met/l2met.go
```diff
@@ -138,4 +138,7 @@ func (p *Provider) log() {
 func (p *Provider) Stop() {}
 
 // Flush implements Provider.
-func (p *Provider) Flush() error { return nil }
+func (p *Provider) Flush() error {
+	p.log()
```
slizco, line 142: nice!

slizco (approved):
LGTM, but one suggestion and a question

cmdutil/service/standard.go
ypaq: The crux is related to this (from [here](https://go.dev/blog/defer-panic-and-recover)):

> Panic is a built-in function that stops the ordinary flow of control and begins panicking. When the function F calls panic, execution of F stops, any deferred functions in F are executed normally, and then F returns to its caller. To the caller, F then behaves like a call to panic. The process continues up the stack until all functions in the current goroutine have returned, at which point the program crashes. 

In my tests, this is the only way the `defer`'d functions get actually called. The program crash happens when [this goroutine](https://github.com/oklog/run/blob/master/group.go#L37-L39) panics where `a.execute` is the call to `sv.Run`. I tried several other variations only this one worked. Happy to be proven wrong though.

slizco: okay, that's fine by me. i see what you mean

cmdutil/service/integration_test.go
ypaq: Checking the log manually was suggestion, as it I couldn't get it to work in a unit test.

agnaite (approved):
Nice!

PR #198, opened 2023-07-18
Make httpmetris work better with OTEL

ypaq:
## Rationale

Here we add a option to improve HTTP metrics reporting with OTEL.

## Changes
* Add `NewV2` function to `httpmetrics` which uses attributes over metric names

## Meta
[W-13781478](https://gus.lightning.force.com/a07EE00001WSFXiYAP)

### DIFF
--- a/hmiddleware/httpmetrics/httpmetrics_v2.go
+++ b/hmiddleware/httpmetrics/httpmetrics_v2.go
@@ -67,7 +67,6 @@ func NewOTEL(p metrics.Provider) func(http.Handler) http.Handler {
 				labels = append(labels, kv...)
 			}
 
-			reg.GetOrRegisterCounter(activeRequests).With(labels...).Add(1)
 			reg.GetOrRegisterExplicitHistogram(requestDuration, metrics.ThirtySecondDistribution).With(labels...).Observe(ms(dur))
 		})
 	}
--- a/hmiddleware/httpmetrics/httpmetrics_v2.go
+++ b/hmiddleware/httpmetrics/httpmetrics_v2.go
@@ -25,11 +25,11 @@ const (
 	urlSchemeKey     = "url.scheme"
 )
 
-// NewV2 returns an HTTP middleware which captures HTTP request counts and latency
+// NewOTEL returns an HTTP middleware which captures HTTP request counts and latency
 // annotated with attributes for method, route, status.
 //
 // See https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/http-metrics/
-func NewV2(p metrics.Provider) func(http.Handler) http.Handler {
+func NewOTEL(p metrics.Provider) func(http.Handler) http.Handler {
 	reg := metricsregistry.New(p)
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
--- a/hmiddleware/httpmetrics/httpmetrics_v2.go
+++ b/hmiddleware/httpmetrics/httpmetrics_v2.go
@@ -5,14 +5,14 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/go-chi/chi"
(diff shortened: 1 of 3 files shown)


hmiddleware/httpmetrics/httpmetrics_v2.go
```diff
@@ -0,0 +1,74 @@
...
+	// metric names
+	requestDuration = "http.server.duration"        // duration in milliseconds
+	activeRequests  = "http.server.active_requests" // counter for number of requests
+
+	// metric attribute keys
+	routeKey         = "http.route"
+	methodKey        = "http.request.method"
+	statusKey        = "http.response.status_code"
+	serverAddressKey = "server.address"
+	urlSchemeKey     = "url.scheme"
```
voidlock, line 25: Nice, this looks it is following guidelines from https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/http-metrics/#http-server

hmiddleware/httpmetrics/httpmetrics_v2.go
```diff
@@ -0,0 +1,74 @@
...
+
+			if r.URL != nil {
+				kv := []string{
+					urlSchemeKey, r.URL.Scheme,
+					serverAddressKey, r.URL.Host,
+				}
+				labels = append(labels, kv...)
+			}
+
+			reg.GetOrRegisterCounter(activeRequests).With(labels...).Add(1)
```
voidlock, line 70: This will change the meaning of [`http.server.active_requests`](https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/http-metrics/#metric-httpserveractive_requests). I know this pattern is matching our current usage with the `http.server.requests` metric, but I suggest tweaking it to match the intended purpose. It should track the active in-flight requests. This will avoid confusion down the road when we standardize on metric naming conventions.

On the line below, we report the metric `http.server.duration` as a histogram. This will automatically make additional metrics available, including:
```
http.server.duration.count
http.server.duration.sum
http.server.duration.avg
http.server.duration.min
http.server.duration.max
http.server.duration.p001...p999
```

Since we will need to update dashboards anyway, I would suggest relying on `http.server.duration.count` instead of our older atomic counter style independent metric.

voidlock (changes requested):
Nice! Once this gets merged, I plan on removing the custom go-logplex version and migrate to on this one.

I left a comment regarding the new `http.server.active_requests` metric though.

hmiddleware/httpmetrics/httpmetrics_v2.go
ypaq: Yes, @slizco pointed this convention out to me. I tried to utilize the [attribute extraction functions](https://pkg.go.dev/go.opentelemetry.io/otel@v1.0.0-RC3/semconv/v1.4.0#HTTPServerMetricAttributesFromHTTPRequest) but they end up adding attributes that would end up with high cardinality.

hmiddleware/httpmetrics/httpmetrics_v2.go
ypaq: Agree, I thought about dropping the `http.server.active_requests` as it is optional anyway. I wondered if `http.server.active_requests` is useful considering that we report on a timer, many values just get dropped which might skew the picture. 

I'm ok with using `http.server.duration.count` as the standard for counting requests.

ypaq: I'm going to drop `http.server.active_requests` for now. We can determine most attributes only when the request is complete. We can rethink this if we need it in the future.

voidlock (approved):

hmiddleware/httpmetrics/httpmetrics_v2.go
voidlock: WFM

ypaq: Thank you

PR #201, opened 2023-10-20
Upgrade to latest opentelemetry-go and use exponential histograms by default

voidlock:
A bunch of opentelemetry related updates:

- Updates the bundled version of opentelemetry-go to the latest stable release.

## Breaking Changes

Older versions of opentelemetry-go used an `ExactHistogram` that exported a metric value for every data point observed. When opentelemetry-go went GA, they removed this type of aggregation. The latest version now offers exponential histogram aggregation. This is more akin to with the histogram support offered via the original `metrics.Provider` offered by this library. You supply the number of buckets, and it just figures out the distributions. 

The alternative is to use an explicit histogram aggregation type where you must specify the boundaries for each bucket. This alternative is less intuitive to use.

## Other Changes

This PR introduces a new endpoint configuration URL, `OTEL_EXPORTER_OTLP_ENDPOINT`. The opentelemetry sdk directly supports this environment variable. The older `OTEL_COLLECTOR_URL` has precedence over the newer environment variable. This ensures backward compatibility and makes transitioning easier.


go-kit/metrics/provider/otel/resource.go
```diff
@@ -0,0 +1,129 @@
...
+	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
+)
+
+// Service level resource attributes
+const (
+	// ServiceKey is the attribute key for the `_service` convention.
+	ServiceKey = attribute.Key("_service")
+
+	// ComponentKey is the attribute key for the `component` convention
+	ComponentKey = attribute.Key("component")
```
ypaq, line 18: Anything for these in semconv?

go-kit/metrics/provider/otel/resource.go
```diff
@@ -0,0 +1,129 @@
...
+// Environment level resource attributes
+const (
+	// StageKey is the attribute key for the `stage` convention.
+	StageKey = attribute.Key("stage")
+
+	// SubServiceKey is the attribute key for the `_subservice` convention.
+	SubServiceKey = attribute.Key("_subservice")
+
+	// CloudKey is the attribute key for the `cloud` convention.
+	CloudKey = attribute.Key("_subservice")
```
ypaq, line 38: Typo here

cmdutil/metrics/otel/otel.go
```diff
@@ -23,24 +20,33 @@ func MustProvider(ctx context.Context, logger logrus.FieldLogger, cfg Config, se
...
 	allOpts := []otel.Option{
-		otel.WithExporter(exporter),
+		// ensure we have service.id, service.namespace, and service.instance.id attributes
+		otel.WithOpenTelemetryStandardService(service, serviceNamespace, serviceInstanceID),
+
+		// ensure we have _service and component attributes
+		otel.WithServiceStandard(service),
+
+		// ensure we have stage and _subcomponent attributes
+		otel.WithEnvironmentStandard(stage),
```
ypaq, line 37: Are we missing the subcomponent argument?

go-kit/metrics/provider/otel/options.go
```diff
@@ -24,140 +23,138 @@ var (
...
-// WithExporter initializes the Provider with an exporter.
-func WithExporter(exp exporter) Option {
-	return func(p *Provider) error {
-		if exp == nil {
-			return ErrExporterNil
+func WithGRPCExporter(endpoint string, options ...otlpmetricgrpc.Option) Option {
+	return WithExporterFunc(func(cfg *config) (metric.Exporter, error) {
+		defaults := []otlpmetricgrpc.Option{
+			otlpmetricgrpc.WithEndpoint(endpoint),
+			otlpmetricgrpc.WithInsecure(),
```
ypaq, line 145: Does it have to be insecure?

go-kit/metrics/provider/otel/provider.go
```diff
@@ -3,371 +3,164 @@ package otel
...
+type viewCache struct {
+	lock    sync.RWMutex
+	streams map[string]metric.Stream
+}
+
+func (v *viewCache) Store(stream metric.Stream) {
+	v.lock.Lock()
+	defer v.lock.Unlock()
+
+	v.streams[stream.Name] = stream
```
ypaq, line 56: Could there be some concurrency issue here because we're not doing a read-before-write?

ypaq (changes requested):
Looks good to me from what I can tell. Requested changes because I believe there are few minor bugs.

go-kit/metrics/provider/otel/resource.go
voidlock: No, these are part of semconv.

go-kit/metrics/provider/otel/options.go
voidlock: This was just copied over from the existing code. https://github.com/heroku/x/pull/201/files/c78ddf68dfd450854cc5ded6c5767429328219eb#diff-b752681e1bce242e23877d7365841220f2745ff1a7bcc49bb4457d0c34faac5aL156-L160

However, I can make it behave the same as the HTTP exporter.

cmdutil/metrics/otel/otel.go
```diff
@@ -5,12 +5,9 @@ import (
 	"strings"
 
 	"github.com/sirupsen/logrus"
-	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric"
-	"go.opentelemetry.io/otel/sdk/export/metric"
 
 	"github.com/heroku/x/go-kit/metrics"
-	"github.com/heroku/x/go-kit/metrics/provider/otel"
+	otel "github.com/heroku/x/go-kit/metrics/provider/otel"
```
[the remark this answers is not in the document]
voidlock: removed

go-kit/metrics/provider/otel/resource.go
voidlock: fixed

go-kit/metrics/provider/otel/options.go
ypaq: Maybe a good opportunity that could avoid surprises in the future.

go-kit/metrics/provider/otel/provider.go
voidlock: I don't think so. There is a higher level lock that happens at the `newHistogram` that would prevent concurrent metric creation calls. There is also a caching layer there that will return an existing histogram if one of the same name has already been created.

cmdutil/metrics/otel/otel.go
voidlock: I think `_subcomponent` is actually a typo from before. I updated this comment to `_subservice`. The option here sets both values, just like before.

go-kit/metrics/provider/otel/options.go
voidlock: I updated the code to match the HTTP exporter. Now `http://` vs `https://` will trigger Insecure vs TLS.

ypaq (approved):
LGTM

PR #204, opened 2024-03-13
Add docs and middleware for dynoid

kennyp:
Rationale
--------------
Dyno identity should be easy to adopt.

Changes
-------------
- Add docs and refactor `dynoid` and `dynoid/dynoidtest`.
- Add http middleware to `dynoid/middleware` to simplify incorporating dynoid.

Meta
--------
[Work Item](https://gus.lightning.force.com/a07EE00001mANddYAG)

### DIFF
--- a/dynoid/dynoid.go
+++ b/dynoid/dynoid.go
@@ -69,8 +69,7 @@ func AllowHerokuHost(host string) IssuerCallback {
 	}
 }
 
-// Subject contains all of the subject information stored by Heroku when issuing
-// a token.
+// Subject contains information about the app and dyno the token was issued for
 type Subject struct {
 	AppID   string `json:"app_id"`
 	AppName string `json:"app_name"`
@@ -120,8 +119,7 @@ func (s *Subject) String() string {
 	return fmt.Sprintf("app:%s.%s::dyno:%s", s.AppID, s.AppName, s.Dyno)
 }
 
-// Subject contains all of the token information stored by Heroku when issuing
-// a token.
+// Token contains all of the token information stored by Heroku when it's issued
 type Token struct {
 	IDToken *oidc.IDToken `json:"-"`
 	SpaceID string        `json:"space_id"`
--- a/dynoid/dynoid.go
+++ b/dynoid/dynoid.go
@@ -8,7 +8,6 @@ import (
 	"log/slog"
 	"os"
 	"path"
-	"strconv"
 	"strings"
 	"sync"
 
@@ -73,17 +72,16 @@ func AllowHerokuHost(host string) IssuerCallback {
 // Subject contains all of the subject information stored by Heroku when issuing
 // a token.
 type Subject struct {
-	AppID      string `json:"app_id"`
-	AppName    string `json:"app_name"`
-	DynoName   string `json:"dyno_name"`
-	DynoNumber int    `json:"dyno_number"`
+	AppID   string `json:"app_id"`
+	AppName string `json:"app_name"`
+	Dyno    string `json:"dyno"`
 }
 
 func (s *Subject) LogValue() slog.Value {
 	return slog.GroupValue(
 		slog.String("app_id", s.AppID),
 		slog.String("app_name", s.AppName),
-		slog.String("dyno", fmt.Sprintf("%s.%d", s.DynoName, s.DynoNumber)),
+		slog.String("dyno", s.Dyno),
 	)
 }
 
@@ -103,20 +101,13 @@ func (s *Subject) UnmarshalText(text []byte) error {
 	}
 
 	app := strings.Split(parts[1], ".")
-	dyno := strings.Split(parts[4], ".")
-
-	if len(app) != 2 || len(dyno) != 2 {
+	if len(app) != 2 {
 		return fmt.Errorf("unexpected subject format: %q", sub)
 	}
 
 	s.AppID = app[0]
 	s.AppName = app[1]
-	s.DynoName = dyno[0]
-	s.DynoNumber, _ = strconv.Atoi(dyno[1])
-
-	if s.DynoNumber == 0 {
-		return fmt.Errorf("unexpected subject format: %q", sub)
-	}
+	s.Dyno = parts[4]
 
 	return nil
 }
@@ -126,7 +117,7 @@ func (s *Subject) String() string {
 		return ""
 	}
 
-	return fmt.Sprintf("app:%s.%s::dyno:%s.%d", s.AppID, s.AppName, s.DynoName, s.DynoNumber)
+	return fmt.Sprintf("app:%s.%s::dyno:%s", s.AppID, s.AppName, s.Dyno)
 }
 
 // Subject contains all of the token information stored by Heroku when issuing
--- a/dynoid/dynoid.go
+++ b/dynoid/dynoid.go
@@ -145,6 +145,7 @@ func (t *Token) LogValue() slog.Value {
 //
 // Suitable for passing as a bearer token
 func ReadLocal(audience string) (string, error) {
+	//nolint: gosec
 	tokenPath := "/etc/heroku/dyno_id_token"
 
 	if audience != "heroku" {
--- a/dynoid/dynoid.go
+++ b/dynoid/dynoid.go
@@ -16,7 +16,10 @@ import (
 )
 
 const (
-	AudienceHeroku = "heroku"
+	DefaultAudience = "heroku"
+
+	//nolint: gosec
+	DefaultTokenPath = "/etc/heroku/dyno_id_token"
 )
 
 // Returned by an IssuerCallback get's an issuer it doesn't trust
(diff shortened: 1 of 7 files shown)


dynoid/dynoid.go
```diff
@@ -5,43 +5,234 @@ import (
...
 	"strings"
 	"sync"
 
 	"github.com/coreos/go-oidc/v3/oidc"
 )
 
-type IssuerCallback func(string) error
+const (
+	AudienceHeroku = "heroku"
+)
```
tt, line 20: Each service is supposed to use a unique audience so defining this as a constant might send the wrong signal.

dynoid/dynoid.go
```diff
@@ -5,43 +5,234 @@ import (
...
 
-type Verifier struct {
-	IssuerCallback IssuerCallback
-	config         *oidc.Config
-	providers      sync.Map
+// Returned by an IssuerCallback get's an issuer it doesn't trust
```
tt, line 22:
```suggestion
// Returned by an IssuerCallback getting an issuer it doesn't trust
```

dynoid/dynoid.go
```diff
@@ -5,43 +5,234 @@ import (
...
+	dyno := strings.Split(parts[4], ".")
+
+	if len(app) != 2 || len(dyno) != 2 {
+		return fmt.Errorf("unexpected subject format: %q", sub)
+	}
+
+	s.AppID = app[0]
+	s.AppName = app[1]
+	s.DynoName = dyno[0]
+	s.DynoNumber, _ = strconv.Atoi(dyno[1])
```
tt, line 112: This is a good example of where we need a later change due to [DR: Deprecating Dyno Numbering on the Heroku Platform](https://salesforce.quip.com/7teIAUvPUQiG) and [RFC: Fir Dyno Naming Separator](https://salesforce.quip.com/SiDbAIjz5w2i).

(Pinging @dianadoherty and @ricardochimal for their awareness.)

dynoid/dynoidtest/dynoidtest.go
```diff
@@ -44,6 +59,10 @@ func (iss *Issuer) GenerateIDToken(clientID string) (string, error) {
 	return token.SignedString(iss.key)
 }
 
+func (iss *Issuer) Context() context.Context {
+	return iss.ctx
+}
+
```
tt, line 65: This feels a bit dirty and seems to only be necessary for the test code. Can we pass a `context.Background()` through there to make the interface smaller?

hmiddleware/dynoid/dynoid.go
```diff
@@ -0,0 +1,121 @@
...
+
+		token, err := verifier.VerifyHeroku(r.Context(), rawToken)
+
+		return r.WithContext(AddToContext(ctx, token, err))
+	}
+}
+
+func tokenFromHeader(r *http.Request) string {
+	bearer := r.Header.Get("Authorization")
+	if len(bearer) > 7 && strings.ToUpper(bearer[0:6]) == "BEARER" {
```
tt, line 117: I suggest matching on the space too and making this case sensitive as recommended in [section 2.1 of RFC 7235](https://datatracker.ietf.org/doc/html/rfc7235#section-2.1):

> It uses a case-insensitive token as a means to identify the authentication scheme, followed by additional information necessary for achieving authentication via that scheme.

```suggestion
	if len(bearer) > 7 && bearer[:7] == "Bearer " {
```

tt (commented):
This looks really good.

I do have one reservation about putting the middleware under the `hmiddleware` package even though it seems like the best fit. This module has quite a few dependencies and I think it would be great if we could eventually break it up a bit.

The dyno ID bits will need to remain together so may consider moving it to `dynoid/dynoidmiddleware` instead? That also prevents two packages that are named identically for import purposes.

dynoid/dynoid.go
kennyp: I was thinking about this after watching @dianadoherty's VAT presentation this morning. Thanks for confirming...

What do you think about just returning `s.Dyno` and not parsing out the name from the number?

dynoid/dynoid.go
kennyp: What if I just don't export the constant?

dynoid/dynoidtest/dynoidtest.go
kennyp: @tt I refactored it a bit to get rid of that. Let me know what you think.

dynoid/dynoid.go
```diff
@@ -5,43 +5,230 @@ import (
...
+
+func (s *Subject) String() string {
+	if s == nil {
+		return ""
+	}
+
+	return fmt.Sprintf("app:%s.%s::dyno:%s", s.AppID, s.AppName, s.Dyno)
+}
+
+// Subject contains all of the token information stored by Heroku when issuing
```
mjrossi, line 123: Did you intend to rewrite this comment to Token specifically, or remove it?

mjrossi (approved):
I think a `readme.md` with some basics on how to use this package would be really helpful. That said, the inline comments definitely help, as does having the tests nearby to use as an example.

dynoid/dynoid.go
kennyp: Good catch! Fixed.

dynoid/dynoid.go
```diff
@@ -56,31 +241,60 @@ func (v *Verifier) Verify(ctx context.Context, rawIDToken string) (*oidc.IDToken
 	return verifier.Verify(ctx, rawIDToken)
 }
 
+// VerifyHeroku verifies the token and parses the returned issuer and subject
+// according to Heroku expected values
+func (v *Verifier) VerifyHeroku(ctx context.Context, rawIDToken string) (*Token, error) {
```
tt, line 246: Considering this is our own package, I think we should merge `Verify` and `VerifyHeroku`.

This seems simpler to maintain so we don't risk someone else starts using it for a use case it wasn't designed for.

dynoid/dynoid.go
```diff
@@ -5,43 +5,228 @@ import (
...
 
-func New(clientID string) *Verifier {
-	return &Verifier{config: &oidc.Config{ClientID: clientID}}
+func (e *UntrustedIssuerError) Error() string {
+	return fmt.Sprintf("untrusted issuer: %v", e.Issuer)
+}
+
+// Returned when the token doesn't match the expected format
+type MalformedTokenError struct {
+	Err error
```
tt, line 35: We already define `Unwrap()` so this doesn't need to be exported:

```suggestion
	err error
```

dynoid/middleware/dynoid.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func internalServerError(format string, v ...any) func(http.Handler) http.Handler {
+	return func(http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprintf(w, format, v...)
+		})
+	}
+}
```
tt, line 122: The formatting is unused:

```suggestion
func internalServerError(error string) func(http.Handler) http.Handler {
	return func(http.Handler) http.Handler {
		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			http.Error(w, error, http.StatusInternalServerError)
		})
	}
}
```

dynoid/middleware/dynoid_test.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+func TestAuthorize(t *testing.T) {
+	router := chi.NewRouter()
+	router.Use(middleware.Authorize(
+		dynoidtest.Audience,
+		dynoid.AllowHerokuHost(dynoidtest.IssuerHost),
+	))
+	router.Get("/", func(w http.ResponseWriter, r *http.Request) {
+	})
```
tt, line 25: This is a nitpick but we don't need the chi dependency. Apologies if there's a syntax error...

```suggestion
	var handler http.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})
	handler = middleware.Authorize(
		dynoidtest.Audience,
		dynoid.AllowHerokuHost(dynoidtest.IssuerHost),
	)(handler)
```

tt (approved):
A few suggestions but still looks good.

siraj-h (approved):

dynoid/dynoid.go
kennyp: Updated in 1910815

dynoid/dynoid.go
kennyp: Fixed in 91aad96

kennyp:
> A few suggestions but still looks good.

Thanks for the suggestions @tt. I've incorporated your feedback.

kennyp:
> I think a `readme.md` with some basics on how to use this package would be really helpful. That said, the inline comments definitely help, as does having the tests nearby to use as an example.

I agree with you @mjrossi, but I figured I would hold off on that until we some devcenter articles and a blog post to link to.

PR #208, opened 2024-05-07
Update newHistogram of testmetrics Provider

Vamsi-Mundra:
**Reason For the Change:** 

While using testmetrics provider in the testcases, we observed CheckObservation is failing when we don't add anything inside the Histogram. 

When we are sending an empty float64 list:  `empty := float64[]` to CheckObservation, it was failing .  

The reason is, we are not initializing observations in the newHistogram function of the testmetrics provider. 

**Changes:** 

Initializing the observations of Histogram in newHistogram function.

### DIFF
--- a/go-kit/metrics/testmetrics/metrics.go
+++ b/go-kit/metrics/testmetrics/metrics.go
@@ -3,7 +3,11 @@ package testmetrics
 import (
 	"sync"
 
+	hll "github.com/axiomhq/hyperloglog"
+
 	"github.com/go-kit/kit/metrics"
+
+	xmetrics "github.com/heroku/x/go-kit/metrics"
 )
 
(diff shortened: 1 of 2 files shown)


ypaq:
Please document next time why this change has the desired effect. From https://pkg.go.dev/reflect#DeepEqual: 
> Slice values are deeply equal when all of the following are true: they are both nil or both non-nil, they have the same length, and either they point to the same initial entry of the same underlying array (that is, &x[0] == &y[0]) or their corresponding elements (up to length) are deeply equal. Note that a non-nil empty slice and a nil slice (for example, []byte{} and []byte(nil)) are not deeply equal.

slizco (approved):

ypaq (approved):

voidlock (approved):

PR #210, opened 2024-05-13
Properly construct auth header for OTEL Metric endpoints

archa347:
Discovered that services using latest version of OTEL utilities from `x` would get authorization failures sending metrics to OTEL collectors that required auth.  Tracked down this bug.


ypaq (approved):
🥇

PR #211, opened 2024-05-15
Update testmetrics's Provider and Metrics

Vamsi-Mundra:
Reason For the Changes

We wanted to update the unit tests of the router library to use the TestMetrics provider of Heroku X. For this, we were required to make some changes in the TestMetrics Provider and metrics.

Even though the current release of Heroku/X is v0.2.0, there are a few issues in sending metrics using that version. Therefore, we want to create a new version, v0.0.62, with the following changes picked from the v0.0.58 branch.

Changes: 

- Added CardinalityCounter which implements CardinalityCounter interface of go-kit/metrics
- Updated Provider to use CardinalityCounter of testmetrics

### DIFF
--- a/go-kit/metrics/testmetrics/metrics.go
+++ b/go-kit/metrics/testmetrics/metrics.go
@@ -3,7 +3,11 @@ package testmetrics
 import (
 	"sync"
 
+	hll "github.com/axiomhq/hyperloglog"
+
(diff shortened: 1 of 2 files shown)


Vamsi-Mundra:
This is not relevant now.

PR #214, opened 2024-07-17
Use RWMutex in OTEL provider

ypaq:
Currently, the `otel.Provider` uses a `sync.Mutex` to protect against race conditions on the maps that store instances of the different metric types. The mutex blocks to ensure the existence of a metric on metric modification. 

We can avoid fully blocking on reads for already existing metrics with a `sync.RWMutex`.

### DIFF
--- a/go-kit/metrics/provider/otel/types.go
+++ b/go-kit/metrics/provider/otel/types.go
@@ -45,10 +45,18 @@ func (p *Provider) NewCounter(name string) metrics.Counter {
 }
 
 func (p *Provider) newCounter(name string, labelValues ...string) metrics.Counter {
+	k := keyName(name, labelValues...)
+
+	p.mu.RLock()
+	c, ok := p.counters[k]
+	p.mu.RUnlock()
+	if ok {
+		return c
+	}
+
 	p.mu.Lock()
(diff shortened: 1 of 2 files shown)


aaronromeo (approved):

agnaite (approved):

slizco (approved):
Looks good to me, but I think a brief description on the PR would be nice. Thanks!

go-kit/metrics/provider/otel/types.go
```diff
@@ -45,10 +45,18 @@ func (p *Provider) NewCounter(name string) metrics.Counter {
 }
 
 func (p *Provider) newCounter(name string, labelValues ...string) metrics.Counter {
+	k := keyName(name, labelValues...)
+
+	p.mu.RLock()
+	c, ok := p.counters[k]
```
hkii, line 51: nit: Not sure if this can be moved to provider.go and get reused.

hkii (approved):

go-kit/metrics/provider/otel/types.go
ypaq: And by "this" you mean the functions in this file?

PR #216, opened 2024-07-26
Set up a pprof server with optional mutex profiling in cmdutil

sherryyaosf:
**Description**

We want to configure the pprof listener that is capable of running mutex profiles on server set-up in `cmdutil/debug`. This PR created a new `NewPprofServer` method to the `cmdutil/debug` package, allowing for the setup of a `pprof` server with configurable profiling types and optional mutex profiling. 

**Changes**
- Add `PProfServerConfig` struct: holds configuration for the pprof server.
- Add `PProfServer` struct: wraps a pprof server
- Implement `NewPProfServer` function: creates and configures the pprof server.
- Implement `Run` and `Stop` methods: the `NewPProfServer` function includes `Run` and `Stop` methods for managing the pprof server lifecycle.
- Write unit tests for `NewPProfServer`

**Metadata**
[W-16252388](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001wqPSIYA2/view)

### DIFF
--- a/cmdutil/debug/debug.go
+++ b/cmdutil/debug/debug.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"net/http"
 	"runtime"
-	"strings"
 	"time"
 
 	"github.com/google/gops/agent"
@@ -131,10 +130,9 @@ func NewPProfServer(config ProfileConfig, l logrus.FieldLogger) *PProfServer {
 // It implements oklog group's runFn.
 func (s *PProfServer) Run() error {
 	s.logger.WithFields(logrus.Fields{
-		"at":       "binding",
-		"service":  "pprof",
-		"addr":     s.addr,
-		"profiles": strings.Join(s.getProfileNames(), ","),
+		"at":      "binding",
+		"service": "pprof",
+		"addr":    s.addr,
 	}).Info()
 
 	if s.pprofServer != nil {
@@ -161,13 +159,4 @@ func (s *PProfServer) Stop(_ error) {
 		}
 	}
 	close(s.done)
-}
-
-// getProfileNames returns the list of profile names configured.
-func (s *PProfServer) getProfileNames() []string {
-	var profiles []string
-	for profile := range s.profileConfig.ProfileHandlers {
-		profiles = append(profiles, profile)
-	}
-	return profiles
 }
--- a/cmdutil/debug/debug.go
+++ b/cmdutil/debug/debug.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/http/pprof"
 	"runtime"
 	"time"
 
@@ -82,17 +83,18 @@ type PProfServer struct {
 	addr          string
 	done          chan struct{}
 	pprofServer   *http.Server
-	profileConfig ProfileConfig
+	profileConfig PProfServerConfig
 }
 
 // ProfileConfig holds the configuration for the pprof server.
-type ProfileConfig struct {
-	Addr            string
-	ProfileHandlers map[string]http.HandlerFunc
+type PProfServerConfig struct {
+	Addr                 string
+	ProfileNames         []string
+	MutexProfileFraction int
 }
 
 // NewPProfServer sets up a pprof server with configurable profiling types and returns a PProfServer instance.
-func NewPProfServer(config ProfileConfig, l logrus.FieldLogger) *PProfServer {
+func NewPProfServer(config PProfServerConfig, l logrus.FieldLogger) *PProfServer {
 	if config.Addr == "" {
 		config.Addr = "127.0.0.1:9998" // Default port
 	}
@@ -101,13 +103,14 @@ func NewPProfServer(config ProfileConfig, l logrus.FieldLogger) *PProfServer {
 	mux := http.NewServeMux()
 
 	// Iterate over the handlers and add them to the mux.
-	for profile, handler := range config.ProfileHandlers {
-		if handler != nil {
-			if profile == "mutex" {
-				runtime.SetMutexProfileFraction(2)
+	for _, profile := range config.ProfileNames {
+		if profile == "mutex" {
+			if config.MutexProfileFraction == 0 {
+				config.MutexProfileFraction = 2 // Use default value of 2 if not set
 			}
-			mux.HandleFunc("/debug/pprof/"+profile, handler)
+			runtime.SetMutexProfileFraction(config.MutexProfileFraction)
 		}
+		mux.Handle("/debug/pprof/"+profile, pprof.Handler(profile))
 	}
 
 	httpServer := &http.Server{
@@ -136,11 +139,9 @@ func (s *PProfServer) Run() error {
 	}).Info()
 
 	if s.pprofServer != nil {
-		go func() {
-			if err := s.pprofServer.ListenAndServe(); err != nil && err != http.ErrServerClosed {
-				s.logger.WithError(err).Error("pprof server error")
-			}
-		}()
+		if err := s.pprofServer.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			return err
+		}
 	}
 
 	<-s.done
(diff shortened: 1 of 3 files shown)


cmdutil/server.go
```diff
@@ -93,3 +97,80 @@ func MultiServer(servers ...Server) Server {
...
+			Handler:           mux,
+			ReadHeaderTimeout: 5 * time.Second,
+		}
+
+		go func() {
+			<-ctx.Done()
+			pprofSrv.Close()
+		}()
+
+		return pprofSrv.ListenAndServe()
```
slizco, line 174: Let's have some logging when we start the server, including what port we are binding to: https://github.com/heroku/x/blob/master/cmdutil/debug/debug.go#L48-L52

slizco (changes requested):

cmdutil/debug/debug.go
```diff
@@ -36,9 +40,10 @@ func New(l logrus.FieldLogger, port int) *Server {
 
 // Server wraps a gops server for easy use with oklog/group.
 type Server struct {
-	logger logrus.FieldLogger
-	addr   string
-	done   chan struct{}
+	logger      logrus.FieldLogger
+	addr        string
+	done        chan struct{}
+	pprofServer *http.Server
```
slizco, line 46: I'd like to see this as a separate server. So something like `debu.PProfServer`.

cmdutil/debug/debug.go
```diff
@@ -70,4 +83,69 @@ func (s *Server) Stop(_ error) {
...
+			l.WithFields(logrus.Fields{
+				"at":      "ignoring",
+				"service": "pprof",
+				"profile": profile,
+			}).Warn("Unknown pprof profile type")
+		}
+	}
+
+	l.WithFields(logrus.Fields{
+		"at":      "binding",
```
slizco, line 133: Binding log should happen when `Run()` is called. The server does not actually start until `Run()`

cmdutil/debug/debug.go
```diff
@@ -70,4 +83,69 @@ func (s *Server) Stop(_ error) {
...
+	}
+
+	// Create a new HTTP mux for handling pprof routes.
+	mux := http.NewServeMux()
+
+	// Iterate over the profile handlers and add them to the mux.
+	for profile, handler := range config.ProfileHandlers {
+		if handler != nil {
+			if profile == "mutex" {
+				runtime.SetMutexProfileFraction(2)
```
slizco, line 115: I think this should be configurable on the profile config.

cmdutil/debug/debug.go
```diff
@@ -70,4 +83,69 @@ func (s *Server) Stop(_ error) {
...
+			}
+			mux.HandleFunc("/debug/pprof/"+profile, handler)
+			l.WithFields(logrus.Fields{
+				"at":      "adding",
+				"service": "pprof",
+				"profile": profile,
+			}).Info("Added pprof profile handler")
+		} else {
```
slizco, line 123: We don't need this else. If someone has passed in a nil-value handler, that is poor configuration we should not support. A nil-pointer exception and subsequent panic are acceptable

cmdutil/debug/debug.go
```diff
@@ -70,4 +83,69 @@ func (s *Server) Stop(_ error) {
...
+		if handler != nil {
+			if profile == "mutex" {
+				runtime.SetMutexProfileFraction(2)
+			}
+			mux.HandleFunc("/debug/pprof/"+profile, handler)
+			l.WithFields(logrus.Fields{
+				"at":      "adding",
+				"service": "pprof",
+				"profile": profile,
+			}).Info("Added pprof profile handler")
```
slizco, line 122: Don't think we need to log this. If we want, we can do a simple `strings.Join()` on the profiles slice and then log `profiles="heap,cpu,blah..."` when we log the port binding in `Run()`.

slizco (changes requested):
I'd also like to see a unit test of this

cmdutil/debug/debug.go
slizco: I mean something like this:
```
type PProfServerConfig struct {
  MutextProfileFraction int
}
```
If set to zero _and_ the mutex profile is enabled, we can use a reasonable default of 2.

cmdutil/debug/debug.go
```diff
@@ -71,3 +75,88 @@ func (s *Server) Stop(_ error) {
...
+	// Create a new HTTP mux for handling pprof routes.
+	mux := http.NewServeMux()
+
+	// Iterate over the handlers and add them to the mux.
+	for profile, handler := range config.ProfileHandlers {
+		if handler != nil {
+			if profile == "mutex" {
+				runtime.SetMutexProfileFraction(2)
+			}
+			mux.HandleFunc("/debug/pprof/"+profile, handler)
```
slizco, line 109: We don't need to raw construction of the handlers like this. The `pprof.Handler` function can construct the handlers for us: https://pkg.go.dev/net/http/pprof#Handler

cmdutil/debug/debug.go
```diff
@@ -71,3 +76,88 @@ func (s *Server) Stop(_ error) {
...
+
+	// Iterate over the handlers and add them to the mux.
+	for _, profile := range config.ProfileNames {
+		if profile == "mutex" {
+			if config.MutexProfileFraction == 0 {
+				config.MutexProfileFraction = 2 // Use default value of 2 if not set
+			}
+			runtime.SetMutexProfileFraction(config.MutexProfileFraction)
+		}
+		mux.Handle("/debug/pprof/"+profile, pprof.Handler(profile))
```
slizco, line 113: Now I am thinking that we should just use `pprof.Index` and generate a handler like so:
```
http.HandlerFunc(pprof.Index)
```
Then you can avoid using mux and having to pass in the list of profiles.

cmdutil/debug/debug.go
```diff
@@ -71,3 +76,88 @@ func (s *Server) Stop(_ error) {
...
+	}
+
+	// Create a new HTTP mux for handling pprof routes.
+	mux := http.NewServeMux()
+
+	// Iterate over the handlers and add them to the mux.
+	for _, profile := range config.ProfileNames {
+		if profile == "mutex" {
+			if config.MutexProfileFraction == 0 {
+				config.MutexProfileFraction = 2 // Use default value of 2 if not set
```
slizco, line 109: Let's not modify the config. That is unexpected behavior. Instead, we can use a local variable:
```
mpf := defaultMutextProfileFraction // value is 2
if config.MutexProfileFraction != 0 {
  mpf = config.MutextProfileFraction
}
runtime.SetMutextProfileFraction(mpf)
```

cmdutil/debug/debug.go
```diff
@@ -71,3 +76,81 @@ func (s *Server) Stop(_ error) {
...
+//
+// It implements oklog group's runFn.
+func (s *PProfServer) Run() error {
+	s.logger.WithFields(logrus.Fields{
+		"at":      "binding",
+		"service": "pprof",
+		"addr":    s.addr,
+	}).Info()
+
+	if s.pprofServer != nil {
```
slizco, line 134: If it is nil, I think we can return an error.

cmdutil/debug/debug_test.go
```diff
@@ -0,0 +1,109 @@
...
+			// Give the server a moment to start
+			time.Sleep(100 * time.Millisecond)
+
+			// Check mutex profile fraction
+			if got := runtime.SetMutexProfileFraction(0); got != tt.expectedMutexFraction {
+				t.Errorf("runtime.SetMutexProfileFraction() = %v, want %v", got, tt.expectedMutexFraction)
+			}
+			runtime.SetMutexProfileFraction(tt.expectedMutexFraction) // Reset to the expected value
+
+			urls := []string{
```
slizco, line 66: I think verifying the paths is probably overkill since that is functionality within `pprof` package.

slizco (approved):
Some minor things but this looks good!

PR #217, opened 2024-08-16
Refactor and Start PProf debug server

Vamsi-Mundra:
**Description**

We want to start the PProf debug server from Standard based on a config

**Changes**
- Updated debug config
- updated NewPProf Server
- Updated standard to start PProf based on EnablePprofDebugging config

**Metadata**
[W-16499473](https://gus.lightning.force.com/a07EE00001z1VZ9YAM)

### DIFF
--- a/cmdutil/debug/config.go
+++ b/cmdutil/debug/config.go
@@ -7,12 +7,7 @@ type Config struct {
 }
 
 type PProfConfig struct {
-	PProfPort            int  `env:"DEBUG_PPROF_PORT,default=9998"`
-	EnablePProfDebugging bool `env:"DEBUG_PPROF_ENABLE,default=false"`
-
-	// Mutex Profiling is not enabled by default
-	EnableMutexProfiling bool `env:"DEBUG_PPROF_MUTEX_PROFILE_ENABLE,default=false"`
-
+	PProfPort int `env:"DEBUG_PPROF_PORT,default=9998"`
 	// This controls how much of fraction of mutexes we need to consider for profiling.
 	MutexProfileFraction int `env:"DBEUG_PPROF_MUTEX_PROFILE_FRACTION,default=2"`
 
@@ -22,10 +17,15 @@ type PProfConfig struct {
 	// Default is actual default value of MemProfileRate
 	MemProfileRate int `env:"DBEUG_PPROF_MEM_PROFILE_RATE,default=524288"`
 
-	// Block Profiling is not enabled default
-	EnableBlockProfiling bool `env:"DEBUG_PPROF_BLOCK_PROFILE_ENABLE,default=false"`
-
 	// This controls how much blocking time in nano seconds we need to consider.
 	// Default 10 indicates, we consider all 10 ns blocking events .
 	BlockProfileRate int `env:"DBEUG_PPROF_BLOCK_PROFILE_RATE,default=10"`
+
+	EnablePProfDebugging bool `env:"DEBUG_PPROF_ENABLE,default=false"`
+
+	// Mutex Profiling is not enabled by default
+	EnableMutexProfiling bool `env:"DEBUG_PPROF_MUTEX_PROFILE_ENABLE,default=false"`
+
+	// Block Profiling is not enabled default
+	EnableBlockProfiling bool `env:"DEBUG_PPROF_BLOCK_PROFILE_ENABLE,default=false"`
 }
--- a/cmdutil/debug/config.go
+++ b/cmdutil/debug/config.go
@@ -3,13 +3,11 @@ package debug
 // Config describes the configurable parameters for debugging.
 type Config struct {
 	Port int `env:"DEBUG_PORT,default=9999"`
-	PProfConfig
+	PProf
 }
 
-type PProfConfig struct {
-	PProfPort int `env:"DEBUG_PPROF_PORT,default=9998"`
-	// This controls how much of fraction of mutexes we need to consider for profiling.
-	MutexProfileFraction int `env:"DBEUG_PPROF_MUTEX_PROFILE_FRACTION,default=2"`
+type PProf struct {
+	Port int `env:"DEBUG_PPROF_PORT,default=9998"`
 
 	// Heap Profiling is enabled by default.
 	// This controls the frequency of heap allocation sampling.
@@ -17,15 +15,18 @@ type PProfConfig struct {
 	// Default is actual default value of MemProfileRate
 	MemProfileRate int `env:"DBEUG_PPROF_MEM_PROFILE_RATE,default=524288"`
 
-	// This controls how much blocking time in nano seconds we need to consider.
-	// Default 10 indicates, we consider all 10 ns blocking events .
-	BlockProfileRate int `env:"DBEUG_PPROF_BLOCK_PROFILE_RATE,default=10"`
-
-	EnablePProfDebugging bool `env:"DEBUG_PPROF_ENABLE,default=false"`
+	Enabled bool `env:"DEBUG_PPROF_ENABLE,default=false"`
 
 	// Mutex Profiling is not enabled by default
 	EnableMutexProfiling bool `env:"DEBUG_PPROF_MUTEX_PROFILE_ENABLE,default=false"`
 
+	// This controls how much of fraction of mutexes we need to consider for profiling.
+	MutexProfileFraction int `env:"DBEUG_PPROF_MUTEX_PROFILE_FRACTION,default=2"`
+
 	// Block Profiling is not enabled default
 	EnableBlockProfiling bool `env:"DEBUG_PPROF_BLOCK_PROFILE_ENABLE,default=false"`
+
+	// This controls how much blocking time in nano seconds we need to consider.
+	// Default 10 indicates, we consider all 10 ns blocking events .
+	BlockProfileRate int `env:"DBEUG_PPROF_BLOCK_PROFILE_RATE,default=10"`
 }
--- a/cmdutil/debug/config.go
+++ b/cmdutil/debug/config.go
@@ -8,25 +8,19 @@ type Config struct {
 
 type PProf struct {
 	Port int `env:"DEBUG_PPROF_PORT,default=9998"`
-
-	// This controls how much of fraction of mutexes we need to consider for profiling.
-	MutexProfileFraction int `env:"DBEUG_PPROF_MUTEX_PROFILE_FRACTION,default=2"`
-
-	// This controls how much blocking time in nano seconds we need to consider.
-	// Default 10 indicates, we consider all 10 ns blocking events .
-	BlockProfileRate int `env:"DBEUG_PPROF_BLOCK_PROFILE_RATE,default=10"`
-
 	// Heap Profiling is enabled by default.
 	// This controls the frequency of heap allocation sampling.
 	// It defines the number of bytes allocated between samples.
 	// Default is actual default value of MemProfileRate
-	MemProfileRate int `env:"DBEUG_PPROF_MEM_PROFILE_RATE,default=524288"`
-
-	Enabled bool `env:"DEBUG_PPROF_ENABLE,default=false"`
-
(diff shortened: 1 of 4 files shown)


cmdutil/debug/config.go
```diff
@@ -2,5 +2,8 @@ package debug
 
 // Config describes the configurable parameters for debugging.
 type Config struct {
-	Port int `env:"DEBUG_PORT,default=9999"`
+	Port                 int  `env:"DEBUG_PORT,default=9999"`
+	PprofPort            int  `env:"PPROF_DEBUG_PORT,default=9998"`
+	EnablePprofDebugging bool `env:"ENABLE_PPROF_DEBUGGING,default=false"`
+	MutexProfileFraction int  `env:"MUTEX_PROFILE_FRACTION,default=2"`
```
ypaq, line 8: We should not enable mutex profiling by default as there is a performance penalty. Could you add an option to enable it?

ypaq (changes requested):

cmdutil/debug/config.go
slizco: That's my bad @ypaq. There is an option above to enable pprof altogether, default to false

cmdutil/debug/config.go
```diff
@@ -2,5 +2,8 @@ package debug
 
 // Config describes the configurable parameters for debugging.
 type Config struct {
-	Port int `env:"DEBUG_PORT,default=9999"`
+	Port                 int  `env:"DEBUG_PORT,default=9999"`
+	PprofPort            int  `env:"PPROF_DEBUG_PORT,default=9998"`
```
slizco, line 6: Can we standardize on the naming of these environment variables? I'm thinking they should all follow the naming convention:
```
DEBUG_PPROF_XXX
```
so you would have:
```
DEBUG_PPROF_ENABLE
DEBUG_PPROF_MUTEX_PROFILE_ENABLE
DBEUG_PPROF_MUTEX_PROFILE_FRACTION
```

cmdutil/service/standard.go
```diff
@@ -74,6 +74,9 @@ func New(appConfig interface{}, ofs ...OptionFunc) *Standard {
 	}
 
 	s.Add(debug.New(logger, sc.Debug.Port))
+	if sc.Debug.EnablePprofDebugging {
+		s.Add(debug.NewPProfServer(logger, sc.Debug.PprofPort, sc.Debug.MutexProfileFraction))
```
slizco, line 78: You could pass these in on a config struct that is embedded on the `debub.Config`:
```
type PprofConfig {
// fields
}
```
and then call `debug.NewPProfServer(sc.Debug.Pprof)` to set up the server.

cmdutil/debug/debug.go
```diff
@@ -85,37 +85,19 @@ type PProfServer struct {
...
-		config.Addr = "127.0.0.1:9998" // Default port
-	}
-
-	// Use a local variable for the mutex profile fraction
-	mpf := defaultMutexProfileFraction
-	if config.MutexProfileFraction != 0 {
-		mpf = config.MutexProfileFraction
-	}
-	runtime.SetMutexProfileFraction(mpf)
+func NewPProfServer(l logrus.FieldLogger, port int, mutexProfileFraction int) *PProfServer {
```
slizco, line 89: Let's standardize on whether or not the second "p" is capitalized. Either `Pprof` or `PProf`

slizco (changes requested):

cmdutil/debug/config.go
```diff
@@ -3,4 +3,29 @@ package debug
 // Config describes the configurable parameters for debugging.
 type Config struct {
 	Port int `env:"DEBUG_PORT,default=9999"`
+	PProfConfig
+}
+
+type PProfConfig struct {
```
ypaq, line 9: The `PProfConfig` is stutter when it's embedded like this. Why not call the embedded config struct `PProf`.

cmdutil/debug/config.go
```diff
@@ -3,4 +3,29 @@ package debug
...
+	// This controls the frequency of heap allocation sampling.
+	// It defines the number of bytes allocated between samples.
+	// Default is actual default value of MemProfileRate
+	MemProfileRate int `env:"DBEUG_PPROF_MEM_PROFILE_RATE,default=524288"`
+
+	// This controls how much blocking time in nano seconds we need to consider.
+	// Default 10 indicates, we consider all 10 ns blocking events .
+	BlockProfileRate int `env:"DBEUG_PPROF_BLOCK_PROFILE_RATE,default=10"`
+
+	EnablePProfDebugging bool `env:"DEBUG_PPROF_ENABLE,default=false"`
```
ypaq, line 24: See my comment above. What do you think about `Config.PProf.Enabled`? The env var can stay the same.

cmdutil/debug/config.go
```diff
@@ -3,4 +3,29 @@ package debug
 // Config describes the configurable parameters for debugging.
 type Config struct {
 	Port int `env:"DEBUG_PORT,default=9999"`
+	PProfConfig
+}
+
+type PProfConfig struct {
+	PProfPort int `env:"DEBUG_PPROF_PORT,default=9998"`
```
ypaq, line 10: This can then be reduced to `Port`.

cmdutil/debug/debug.go
```diff
@@ -135,6 +126,7 @@ func (s *PProfServer) Run() error {
 		"addr":    s.addr,
 	}).Info()
 
+	// runtime.MemProfileRecord
```
ypaq, line 129: Remove commented code.

ypaq (changes requested):

cmdutil/debug/config.go
```diff
@@ -3,4 +3,29 @@ package debug
...
+	// Default 10 indicates, we consider all 10 ns blocking events .
+	BlockProfileRate int `env:"DBEUG_PPROF_BLOCK_PROFILE_RATE,default=10"`
+
+	EnablePProfDebugging bool `env:"DEBUG_PPROF_ENABLE,default=false"`
+
+	// Mutex Profiling is not enabled by default
+	EnableMutexProfiling bool `env:"DEBUG_PPROF_MUTEX_PROFILE_ENABLE,default=false"`
+
+	// Block Profiling is not enabled default
+	EnableBlockProfiling bool `env:"DEBUG_PPROF_BLOCK_PROFILE_ENABLE,default=false"`
```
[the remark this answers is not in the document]
Vamsi-Mundra: We are getting lint error when we are grouping them
https://github.com/heroku/x/actions/runs/10565031406/job/29268699927?pr=217#step:4:66

cmdutil/service/standard.go
```diff
@@ -74,6 +74,9 @@ func New(appConfig interface{}, ofs ...OptionFunc) *Standard {
 	}
 
 	s.Add(debug.New(logger, sc.Debug.Port))
+	if sc.Debug.Enabled {
+		s.Add(debug.NewPProfServer(logger, &sc.Debug.PProf))
+	}
```
ypaq, line 79: Why don't we do this inside `debug.New`?

cmdutil/debug/debug.go
```diff
@@ -85,37 +85,28 @@ type PProfServer struct {
 	pprofServer *http.Server
 }
 
-// ProfileConfig holds the configuration for the pprof server.
-type PProfServerConfig struct {
-	Addr                 string
-	MutexProfileFraction int
-}
+// NewPProfServer sets up a pprof server with configurable profiling types and returns a PProfServer instance.
+func NewPProfServer(l logrus.FieldLogger, pprofConfig *PProf) *PProfServer {
```
ypaq, line 89: Why do we need to create a new server. Why can't this be part of the `debug.Server`?

ypaq (changes requested):

cmdutil/debug/debug.go
```diff
@@ -31,25 +33,72 @@ import (
...
+func New(l logrus.FieldLogger, config Config) *Server {
+	server := &Server{
 		logger: l,
-		addr:   fmt.Sprintf("127.0.0.1:%d", port),
+		addr:   fmt.Sprintf("127.0.0.1:%d", config.Port),
 		done:   make(chan struct{}),
 	}
+	if config.Enabled {
+		server.pprof = NewPProfServer(l, &config.PProf)
+	}
```
ypaq, line 44: We're duplicating a server here.

cmdutil/debug/debug.go
```diff
@@ -31,25 +33,72 @@ import (
...
 
 // Run starts the debug server.
 //
-// It implements oklog group's runFn.
+// It implements oklog group's runFn and pprof.
 func (s *Server) Run() error {
+
+	var wg sync.WaitGroup
+	gopsErrChan := make(chan error, 1)
+	pprofErrChan := make(chan error, 1)
```
ypaq, line 63: Do we need two done channels?

ypaq (approved):
Left some comments

cmdutil/debug/debug.go
Vamsi-Mundra: @ypaq Are you suggesting that we can do it in this way?
https://github.com/heroku/x/pull/218 ?

cmdutil/debug/debug.go
Vamsi-Mundra: We are starting two servers one for gops agent and one for pprof. 
So I have added two error channels for the two go routines

slizco: You don't need these error channels at all because you are using the `sync.WaitGroup`.

cmdutil/debug/debug.go
```diff
@@ -31,25 +33,74 @@ import (
...
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		gopsErrChan <- s.RunGOPS()
```
slizco, line 70: Why should this be a channel? You can simply do:
```
gopsErr = s.RunGOPS()
```

The call to `wg.Wait()` on line 81 will mean that you block until both services have stopped running. In fact, you should probably catch the error here and log it if one of the services has stopped running. You could end up with the pprof server exiting in error and gops server running indefinitely, with no indication as to why or that this happened.

slizco (changes requested):

cmdutil/debug/config.go
```diff
@@ -3,4 +3,30 @@ package debug
...
+
+type PProf struct {
+	Port int `env:"DEBUG_PPROF_PORT,default=9998"`
+
+	// This controls how much of fraction of mutexes we need to consider for profiling.
+	MutexProfileFraction int `env:"DBEUG_PPROF_MUTEX_PROFILE_FRACTION,default=2"`
+
+	// This controls how much blocking time in nano seconds we need to consider.
+	// Default 10 indicates, we consider all 10 ns blocking events .
+	BlockProfileRate int `env:"DBEUG_PPROF_BLOCK_PROFILE_RATE,default=10"`
```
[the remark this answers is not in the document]
Vamsi-Mundra: I have removed the lines, is the current state you suggested

cmdutil/debug/debug.go
Vamsi-Mundra: Updated it

cmdutil/debug/config.go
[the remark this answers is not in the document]
slizco: I suppose any lines with comments in between prevents the auto-formatter from aligning. Feel free to format this as you wish

cmdutil/debug/debug.go
```diff
@@ -31,25 +33,75 @@ import (
...
+		}
+	}()
+
+	if s.pprof != nil {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			pprofErr = s.pprof.Run()
+			if pprofErr != nil {
+				s.pprof.logger.WithError(gopsErr).Error("pprof server failed")
```
slizco, line 82: We're logging the wrong error here.

PR #219, opened 2024-10-01
go-kit: Add CheckCounterExists for testmetrics

Vamsi-Mundra:
Adding a `CheckCounterExists` function to `testmetrics` would be helpful for scenarios where we need to verify that a metric exists and is not 0, but we can't determine the exact counter value.

### DIFF
--- a/go-kit/metrics/testmetrics/provider.go
+++ b/go-kit/metrics/testmetrics/provider.go
@@ -171,7 +171,7 @@ func (p *Provider) CheckNoCounter(name string, labelValues ...string) {
 }
 
 // CheckCounterExists checks that there is registered counter with the name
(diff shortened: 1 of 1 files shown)


go-kit/metrics/testmetrics/provider.go
```diff
@@ -170,6 +170,24 @@ func (p *Provider) CheckNoCounter(name string, labelValues ...string) {
 	}
 }
 
+// CheckCounterExists checks that there is registered counter with the name
+// provided and value is greater than 0.
```
slizco, line 174: Technically, your function checks that the value _is not zero_ which is different from _greater than 0_. Let's fix the logic or update the comment to match.

slizco (approved):

PR #221, opened 2024-10-03
Add ability to override where tokens are read from.

kennyp:
## Rationale
- Allows for testing code that reads the token.
- Adds missing docs.

## Links
- [Work Item](https://gus.lightning.force.com/a07EE000022Vuv7YAC)


mjrossi (approved):

dynoid/dynoid.go
```diff
@@ -142,11 +143,28 @@ func LocalTokenPath(audience string) string {
...
+func (*osReader) ReadFile(name string) ([]byte, error) {
+	return os.ReadFile(name)
+}
+
+// DefaultFS is used by [ReadLocal] and [ReadLocalToken] to retrieve tokens.
+//
+// By default they are retrieved via [os.Open] and [os.ReadFile].
+//
+// This is useful when testing code that uses DynoID.
+var DefaultFS fs.ReadFileFS = &osReader{}
```
tt, line 161: Would `LocalFS` maybe be better to match the method naming?

tt (approved):

dynoid/dynoid.go
kennyp: I was just following the convention from other Go packages in the stdlib. But then again they do have a set of methods that works on the default and another set that allows you to instantiate or pass in your own. :shrug: 

I think I'll leave it for now and we can re-visit as we play with it.

PR #223, opened 2024-10-04
Fix Warnings and Lint Errors

kennyp:
## Rationale
- Fix gRPC warnings.
- Remove unused arguments.
- Don't shadow built-in functions.
- Don't set PreferServerCipherSuites since it's ignored.
- Safe type conversions.

## Links
- Fixes: https://github.com/heroku/x/security/code-scanning/7


dynoid/dynoid.go
```diff
@@ -90,7 +90,7 @@ func (s *Subject) MarshalText() ([]byte, error) {
 
 func (s *Subject) UnmarshalText(text []byte) error {
 	if s == nil {
-		*s = Subject{}
+		return fmt.Errorf("cannont unmarshal to nil")
```
tt, line 93:
```suggestion
		return fmt.Errorf("cannot unmarshal to nil")
```

tt (approved):

PR #225, opened 2024-10-14
DynoID Helpers for Testing and Local Development

kennyp:
## Rationale
- Add `GenerateDefaultFS` function to populate `DefaultFS` using a local issuer.
- Add `LocalIssuer` middleware to configure the oidc context for incoming requests.
- Add `Issue` handler to provide an endpoint for minting tokens using the local issuer.
- Add `LocalConfiguration` and `ConfigureLocal` for one-step development/test configuration of DynoID.
- Add `WithKey` IssuerOpt for providing a private key.
- Handle `nil` when logging.

## Links
- [Work Item](https://gus.lightning.force.com/a07EE000023Ie9wYAC)


mjrossi (commented):
I did some testing with this, and did get something working that generates a token for local development, and if I use that token for requests it does allow auth.

However, I noticed that it did _not_ work with the helper `AuthorizeSpaces` -- I believe specifically `dynoid.AllowHerokuSpace` in that method definition will trigger the use of the default issuer that assumes whatever host you pass it -- this was the discussion we had yesterday around specifying `heroku.com` if you recall. 

If I changed my local code to use `AuthorizeSameSpace`, it worked. I do need to authorize to multiple spaces though, so I think we need one more helper to do the trick or I can make it on my side.

```go
// first I have this
if os.Getenv("ENVIRONMENT") == "local" {
		cfg, err := dynoidtest.ConfigureLocal([]string{DEFAULT_DYNOID_AUDIENCE})
		if err != nil {
			slog.Error("error configuring local dynoid", slog.Any("error", err))
		}

		token, err := cfg.GenerateToken(DEFAULT_DYNOID_AUDIENCE)
		if err != nil {
			slog.Error("error generating local dynoid token", slog.Any("error", err))
		}

		fmt.Printf("TOKEN FOR LOCAL REQUESTS: %s\n", token)

		router.Use(cfg.Middleware())
	}

	router.Use(authorizeDynoID(DEFAULT_DYNOID_AUDIENCE))

// inside of the authorizeDynoID which is a function I created, I call these methods from this package after verifying the excluding route

// this one worked with my other setup
       authMiddleware := dynoidmiddleware.AuthorizeSameSpace(audience)
// this one does not work
       authMiddleware := dynoidmiddleware.AuthorizeSpaces(audience, "heroku.com", authorizedSpaces...)
```

Getting there! 🙌

kennyp:
> I did some testing with this, and did get something working that generates a token for local development, and if I use that token for requests it does allow auth.
> 
> However, I noticed that it did _not_ work with the helper `AuthorizeSpaces` -- I believe specifically `dynoid.AllowHerokuSpace` in that method definition will trigger the use of the default issuer that assumes whatever host you pass it -- this was the discussion we had yesterday around specifying `heroku.com` if you recall.
> 
> If I changed my local code to use `AuthorizeSameSpace`, it worked. I do need to authorize to multiple spaces though, so I think we need one more helper to do the trick or I can make it on my side.
> 
> ```go
> // first I have this
> if os.Getenv("ENVIRONMENT") == "local" {
> 		cfg, err := dynoidtest.ConfigureLocal([]string{DEFAULT_DYNOID_AUDIENCE})
> 		if err != nil {
> 			slog.Error("error configuring local dynoid", slog.Any("error", err))
> 		}
> 
> 		token, err := cfg.GenerateToken(DEFAULT_DYNOID_AUDIENCE)
> 		if err != nil {
> 			slog.Error("error generating local dynoid token", slog.Any("error", err))
> 		}
> 
> 		fmt.Printf("TOKEN FOR LOCAL REQUESTS: %s\n", token)
> 
> 		router.Use(cfg.Middleware())
> 	}
> 
> 	router.Use(authorizeDynoID(DEFAULT_DYNOID_AUDIENCE))
> 
> // inside of the authorizeDynoID which is a function I created, I call these methods from this package after verifying the excluding route
> 
> // this one worked with my other setup
>        authMiddleware := dynoidmiddleware.AuthorizeSameSpace(audience)
> // this one does not work
>        authMiddleware := dynoidmiddleware.AuthorizeSpaces(audience, "heroku.com", authorizedSpaces...)
> ```
> 
> Getting there! 🙌

@mjrossi  the easiest fix there is to replace the hardcoded "heroku.com" with a call to a function. (Which you could cache off or something).

```golang
func host() string {
	if os.Getenv("environment") == "local" {
		return dynotest.IssuerHost
	}

	return "heroku.com"
}
```

But generally speaking are you thinking we should have a middleware that leaves out the host param and assumes same issuer as the current app?

mjrossi:
@kennyp yeah, for sure -- the way I've architected stuff right now will make two branches to an `os.Getenv(...)` call, which isn't a big deal but was trying to simplify.

> But generally speaking are you thinking we should have a middleware that leaves out the host param and assumes same issuer as the current app?

Yeah, exactly this.

hsophie-sf:
Example snippets of how to use the library and the local token would be helpful. I'm still having difficulty wrapping my head around this.

mjrossi:
@kennyp as you know the current state of this works great for me. Will review again once you mark as ready 👍

Makefile
```diff
@@ -78,3 +78,10 @@ proto: $(TOOLS_BIN)/protoc $(TOOLS_BIN)/protoc-gen-go | $(TOOLS_BIN) ## Regenera
 		--go_out=paths=source_relative:. \
 		--loggingtags_out=. \
 		./cmd/protoc-gen-loggingtags/internal/test/*.proto
+
+$(GOPATH)/bin/gomarkdoc:
+	go install github.com/princjef/gomarkdoc/cmd/gomarkdoc@latest
+
+.PHONY: docs
+docs: $(GOPATH)/bin/gomarkdoc ## Generate docs using gomarkdoc
+	$< --header="" -o ./dynoid/README.md -e ./dynoid/...
```
tt, line 87: Should there be a CI step to ensure you ran this?

dynoid/README.md
```diff
@@ -0,0 +1,884 @@
+> [!IMPORTANT]
+> DynoID is currently a [Heroku Labs][labs] feature and is not enabled by default for all spaces.
+
+# DynoID
+
+DynoID is a feature of [Heroku Private Spaces][spaces] that leverages [OpenID Connect][oidc] to mint tokens for each dyno on boot that it can use to authenticate and verify their identity to other services.
```
tt, line 6: I suggest removing "on boot" as they're also renewed periodically:

```suggestion
DynoID is a feature of [Heroku Private Spaces][spaces] that leverages [OpenID Connect][oidc] to mint tokens for each dyno that it can use to authenticate and verify their identity to other services.
```

dynoid/README.md
```diff
@@ -0,0 +1,884 @@
...
+
+## Audiences
+
+All dynos are get an `heroku` audience token by default. Additional audiences can be minted by setting the `HEROKU_DYNO_ID_AUDIENCES` config var to a comma separated list of audiences.
```
tt, line 10:
```suggestion
All dynos get a `heroku` audience token by default. Additional audiences can be minted by setting the `HEROKU_DYNO_ID_AUDIENCES` config var to a comma separated list of audiences.
```

dynoid/README.md
```diff
@@ -0,0 +1,884 @@
...
+
+## To Authenticate Calls to Your Service
+
+The [dynoid](<#dynoid>) package provides all of the functions needed to verify a token issued for an app in an Heroku Space. Additionally there is a [middleware](<#middleware>) package that provides a set of http [handlers][handler] and middleware suitable for use in a web application.
```
tt, line 14:
```suggestion
The [dynoid](<#dynoid>) package provides all of the functions needed to verify a token issued for an app in a Heroku Private Space. Additionally there is a [middleware](<#middleware>) package that provides a set of http [handlers][handler] and middleware suitable for use in a web application.
```

dynoid/dynoid.go
```diff
@@ -60,9 +60,9 @@ type IssuerCallback func(issuer string) error
 
 // AllowHerokuHost verifies that the issuer is from Heroku for the given host
 // domain
-func AllowHerokuHost(host string) IssuerCallback {
+func AllowHerokuHost(issuerHost string) IssuerCallback {
```
tt, line 63: I think this change is a bit misleading as it's not the issuer host but the `HEROKU_HOST` equivalent you're providing.

dynoid/dynoidtest/dynoidtest.go
```diff
@@ -21,18 +21,17 @@ import (
...
 
 const (
-	// IssuerHost is the host used by the dynoidtest.Issuer
-	IssuerHost = "heroku.local"
-
-	DefaultSpaceID = "test"                                 // space id used when one is not provided
-	DefaultAppID   = "00000000-0000-0000-0000-000000000001" // app id used when one is not provided
-	DefaultAppName = "sushi"                                // app name used when one is not provided
-	DefaultDyno    = "web.1"                                // dyno used when one is not provided
+	DefaultIssuerHost = "heroku.local"                         // issuer host used when one is not provided
```
tt, line 24: This was misleading before. The issuer host is "oidc.heroku.local" so this should probably be the Heroku host:

```suggestion
	DefaultHerokuHost = "heroku.local"                         // issuer host used when one is not provided
```

dynoid/middleware/example_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"io"
+	"log"
+	"net/http"
+
+	"github.com/heroku/x/dynoid/middleware"
+)
+
+const AUDIENCE = "testing"
```
tt, line 11: I think [constants should be mixed caps](https://google.github.io/styleguide/go/decisions.html#constant-names) although maybe this is to signal that you replace it?

Makefile
kennyp: Good call!

kennyp:
@tt Verifying the docs works fine locally, but fails on CI. Any ideas?

kennyp:
> @tt Verifying the docs works fine locally, but fails on CI. Any ideas?

Never mind, fixed it. Looks like it can't figure out repo stuff without a little help when running in an action.

hsophie-sf (approved):

mjrossi (approved):

.github/workflows/ci.yaml
```diff
@@ -5,9 +5,11 @@ on: [ push ]
 jobs:
   ci:
     runs-on: [ ubuntu-latest ]
+    env:
+      GOMARKDOC_OPTS: -c
     strategy:
       matrix:
-        target: [ test, lint ]
+        target: [ test, lint, docs ]
```
tt, line 12: Does this simply generate the documentation or actually verify that the generated content is identical to what's in the repository?

Makefile
```diff
@@ -78,3 +79,11 @@ proto: $(TOOLS_BIN)/protoc $(TOOLS_BIN)/protoc-gen-go | $(TOOLS_BIN) ## Regenera
...
 		--loggingtags_out=. \
 		./cmd/protoc-gen-loggingtags/internal/test/*.proto
+
+$(GOPATH)/bin/gomarkdoc:
+	go install github.com/princjef/gomarkdoc/cmd/gomarkdoc@latest
+
+.PHONY: docs
+docs: $(GOPATH)/bin/gomarkdoc ## Generate docs using gomarkdoc
+	$< $(GOMARKDOC_OPTS) -o ./dynoid/README.md -e ./dynoid/...
+
```
tt, line 89:
```suggestion
```

dynoid/dynoid.go
```diff
@@ -187,10 +195,10 @@ func ReadLocalToken(ctx context.Context, audience string) (*Token, error) {
 
 // AllowHerokuSpace verifies that the issuer is from Heroku for the given host
 // and space id.
-func AllowHerokuSpace(host string, spaceIDs ...string) IssuerCallback {
+func AllowHerokuSpace(issuerHost string, spaceIDs ...string) IssuerCallback {
```
tt, line 198: This is also the `HEROKU_HOST`:

```suggestion
func AllowHerokuSpace(herokuHost string, spaceIDs ...string) IssuerCallback {
```

tt (approved):

.github/workflows/ci.yaml
tt: Oh, the `GOMARKDOC_OPTS=-c` does that?

I guess it's fine but could be more discoverable to still have a `verify-docs` target and set the environment there. I don't feel strongly about this, though.

PR #224, opened 2024-10-10
Update runs-on to allow for CodeQL-Build to run.

kennyp:

PR #226, opened 2024-10-15
override default token to avoid hitting ip allowlist

sheax0r:


kennyp:
I think this was fixed when I merged #224

PR #228, opened 2024-12-05
Add Option to wait while stopping Server

Vamsi-Mundra:

Currently there is option to wait before shutting down server when we receive an interrupt Signal. We want to add wait time to signals server,  so that other severs running using the standard server will have time to do any shutdown shenanigans.

## Description of changes

- added SIGNALS_SERVER_STOP_DELAY with default 0
- added signals config
- added wait insider signals server
- update standardConfig.

[W-16790435](https://gus.lightning.force.com/a07EE000021ZVpgYAG)

### DIFF
--- a/cmdutil/signals/config.go
+++ b/cmdutil/signals/config.go
@@ -1,6 +1,8 @@
 package signals
 
+import "time"
+
 // Config describes the configurable parameters for Signals server.
 type Config struct {
-	ServerCloseWaitTime int `env:"SERVER_CLOSE_WAIT_TIME,default=0"`
+	SignalsServerStopDelay time.Duration `env:"SIGNALS_SERVER_STOP_DELAY,default=0s"`
 }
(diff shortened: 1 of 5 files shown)


cmdutil/signals/config.go
```diff
@@ -0,0 +1,6 @@
+package signals
+
+// Config describes the configurable parameters for Signals server.
+type Config struct {
+	ServerCloseWaitTime int `env:"SERVER_CLOSE_WAIT_TIME,default=0"`
```
ypaq, line 5: Maybe we should namespace the environment variable and make its purpose more explicit, for example, `SIGNALS_SERVER_STOP_DELAY`.

The type should be `time.Duration` and the default `0s`.

cmdutil/signals/signals.go
```diff
@@ -43,6 +44,7 @@ func NewServer(logger logrus.FieldLogger, signals ...os.Signal) cmdutil.Server {
 			if sig != nil {
 				logger.Infoln("received signal", sig)
 			}
+			time.Sleep(time.Duration(config.ServerCloseWaitTime) * time.Second)
```
ypaq, line 47: You'd have to update this accordingly.

ypaq (changes requested):

PR #280, opened 2026-04-01
Add flexible base resource option to otel provider

alexmarnell:
Eliminates schema URL version conflicts by using schemaless resources throughout the provider initialization, then applying a schema at the end only if the user hasn't provided one.

## Problem

Applications upgrading their OpenTelemetry SDK encounter schema URL conflicts:
```
failed to merge resources: conflicting Schema URL: 
https://opentelemetry.io/schemas/1.39.0 and https://opentelemetry.io/schemas/1.37.0
```

This occurs when downstream applications use a different OTel SDK version than this library.

## Solution

Uses schemaless resources to avoid conflicts during merging:

1. Start with schemaless base (environment detection without schema)
2. All helper functions create schemaless resources
3. All merging happens without schemas (no conflicts possible)
4. At the end, apply library's schema only if user didn't provide one

**Changes:**
- All resource helpers now use `resource.NewSchemaless()`
- Provider starts with schemaless base from `resource.Default()`
- Final check adds library's schema (v1.37.0) if missing

## Backward Compatibility

Existing code works unchanged - uses library's v1.37.0 schema by default.

## Upgrade Path

Users with newer OTel SDK versions can provide their own schema:
```go
import semconv "go.opentelemetry.io/otel/semconv/v1.39.0"

base := resource.NewWithAttributes(semconv.SchemaURL)
p, err := otel.New(ctx, "my-service",
    otel.WithResource(base),
    otel.WithOpenTelemetryStandardService("my-service", "my-ns", "inst-1"),
)
```

All helpers merge schemaless attributes while preserving the user's schema version.

### DIFF
--- a/go-kit/metrics/provider/otel/provider.go
+++ b/go-kit/metrics/provider/otel/provider.go
@@ -10,6 +10,8 @@ import (
 	"go.opentelemetry.io/otel/sdk/metric"
 	"go.opentelemetry.io/otel/sdk/resource"
 
+	semconv "go.opentelemetry.io/otel/semconv/v1.37.0"
+
 	xmetrics "github.com/heroku/x/go-kit/metrics"
 )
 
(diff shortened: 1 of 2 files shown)


ypaq (approved):
LGTM

